PR #25, opened 2019-05-16
Explicitly require PHP 7+

danepowell:
ORCA uses some language constructs that are only available in PHP 7+, so it seems like it should explicitly declare this dependency.


TravisCarden:
Quite so, @danepowell. Thanks!

PR #28, opened 2019-07-03
Configure phpstan to ignore deprecation of AnnotationsRegistry

### LINKED ISSUE #27: Ignore deprecation of AnnotationsRegistry
ORCA scans for deprecated function calls by default.

BLT uses ORCA as well as the AnnotationsRegistry from doctrine/annotations. Unfortunately, AnnotationsRegistry was deprecated even though no alternative currently exists (an alternative will come in the next major release). I consider this a bug in the annotations library, but the maintainer hasn't expressed willingness to fix the situation: https://github.com/doctrine/annotations/issues/270

In order to get deprecation tests running again, I think the best solution would be for ORCA to add a phpstan `ignoreErrors` directive to ignore the following deprecation error until the upstream bug with doctrine/annotations is fixed:
`Call to deprecated method registerFile() of class Doctrine\Common\Annotations\AnnotationRegistry.`

If there were any way to solve this within BLT without involving ORCA I definitely would, but it looks like the only way phstan can ignore false positives is via global configuration: https://github.com/phpstan/phpstan/issues/786

TravisCarden:


TravisCarden:
The Pipelines failure is unrelated to the PR and will be addressed separately.

PR #30, opened 2019-07-09
Fixed false positives due to system exec calls.

danepowell:
Adds fix for upstream issue: https://github.com/FloeDesignTechnologies/phpcs-security-audit/issues/42


TravisCarden:
Thanks, @danepowell!

PR #32, opened 2019-07-12
Revising Behat Example and Docs.

mikemadison13:
This update is based on the real life example i see in https://github.com/acquia/lightning


mikemadison13:
odd @TravisCarden, let me keep poking at that. that methodology definitely worked for me in my distro (and it seems to be working in lightning).

mikemadison13:
also @TravisCarden this may have broken because of composer caches. i had issues with the new composer autoloading i did until i flushed the travis cache.

mikemadison13:
note that i accidentally left the `paths` key out of defaults in the original PR so it wasn't autoloading in the feature context properly. appears to be working now

TravisCarden:
Thanks, @mikemadison13! Merged.

PR #33, opened 2019-07-16
Fix potential Phpstan autoloader conflicts leading to incorrect deprecation scan results

TravisCarden:
Because Phpstan loads the autoloader in its parent `vendor/` directory before the autoloader file specified on the command-line, ORCA's vendor libraries will be preferred over those in the test fixture when there is a conflict, potentially leading to incorrect deprecation scan results (c.f. https://github.com/acquia/blt/pull/3712#issuecomment-509834818). This fixes the problem by installing Phpstan to and running it from the fixture instead.


danepowell:
Nice fix. I'll work on resolving the downstream BLT issue ASAP so you can merge this.

danepowell:
I think this can be merged now, right? I know there was talk last week of requiring `composer/composer` as well, but that caused side effects, and seems like it should go in a separate issue.

TravisCarden:
Merged. Thanks, @danepowell.

PR #34, opened 2019-07-17
Change SUT repository from "path" to "package" type for more robustness when developing locally

TravisCarden:
Composer path repositories work well for placing the SUT on Travis CI where the checked out Git branch gets renamed to match the target branch (i.e., `ORCA_SUT_BRANCH`), but developing locally with arbitrary topic branches checked out branch can lead to unexpected failures like "Could not find package drupal/example in a version matching dev-my-topic-branch". A package repository using whatever Composer considers the local repo's version supports this use case without changing the behavior on Travis CI.


TravisCarden:
This change has been reverted because it broke Composer dependency resolution in several initially undetected ways.

PR #55, opened 2019-12-12
Added Composer Scaffold to ORCA

danepowell:
https://github.com/acquia/composer-scaffold is a minimal Composer meta-package that just provides Composer Scaffold settings that are specific to Acquia, especially setting the web-root to `docroot`. In order for it to be made open-source, it must integrate with ORCA.

It seems there's a bit of a catch-22 here, because ORCA can't composer require it until it's open source. So for now, I'm applying this PR as a patch in https://github.com/acquia/composer-scaffold/pull/2

cc @jfarrell

### DIFF
--- a/config/packages.yml
+++ b/config/packages.yml
@@ -62,7 +62,8 @@ acquia/coding-standards:
   type: phpcodesniffer-standard
   url: ../coding-standards-php
   
-acquia/composer-scaffold: []
+acquia/composer-scaffold:
+  type: package
 
 drupal/cog:
   type: drupal-theme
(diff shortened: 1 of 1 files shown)


danepowell:
I was mistaken, the composer-scaffold repo is open source. There must be some other reason ORCA can't find it.

@TravisCarden when I run `composer require acquia/composer-scaffold:"dev-master"` locally it works fine, but the ORCA tests above are failing with `[ERROR] No available version could be found for acquia/composer-scaffold:"dev-master"`. Any thoughts?

danepowell:
Bah... I discovered that some of the delegation abilities related to Composer Scaffold are more limited than I thought, so there's not as much value to meta packages. BLT is going to stop relying on this for instance. So not much point testing it.

PR #62, opened 2020-02-24
Add BLT's ci.settings.php file to fixtures

TravisCarden:
Closes #64

### DIFF
--- a/src/Fixture/FixtureCreator.php
+++ b/src/Fixture/FixtureCreator.php
@@ -980,23 +980,28 @@ class FixtureCreator {
    */
   protected function ensureDrupalSettings(): void {
     $this->output->section('Ensuring Drupal settings');
-    $this->ensureSettingsFile('ci.settings.php');
-    $this->ensureSettingsFile('local.settings.php');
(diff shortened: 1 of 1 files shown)


danepowell:
I'm stumped. Generally, the error appears to be that the db credentials are invalid, but I'm not sure why this PR would have changed that. I'll have to dig in.

danepowell:
Ah, now I understand.

BLT automatically sets `$databases['default']['default']` on Pipelines to be able to access MySQL with the default credentials: https://github.com/acquia/blt/blob/12.x/settings/pipelines.settings.php

However it will happily override this with anything that you specify in a more specific `ci.settings.php` such as you've created in this PR.

The problem then is that you are appending to the default local settings file which specifies "default" db credentials of drupal/drupal/drupal, overriding the correct credentials for Pipelines.

There's a few ways you could fix this, but I think the most correct is to ensure that local.settings.php and ci.settings.php don't set any db credentials unless you are overriding the default to use SQLite.

danepowell:
@TravisCarden I took a crack at fixing this for you in #64, let me know what you think.

PR #65, opened 2020-03-06
Fix local path repo for blt-project.

danepowell:
BLT manages BLT Project as a subtree split in its repo, and any changes to BLT Project need to be tested when BLT is the SUT. This means pulling in acquia/blt-project as a Composer path repo.

The problem is that Composer is incredibly finicky about how it grabs (or doesn't grab) local versions of packages. Right now BLT copies blt-project to `/tmp/blt-project` as part of its test setup, and then ORCA tries to include it with `composer create-project acquia/blt-project:12.x` (or 11.x). This fails because the tmp copy of blt-project has no version info, so Composer assumes it to be `dev-master` and goes looking for an online copy when you ask for 12.x.

The best solution I can think of is for ORCA to ask for `dev-master` whenever BLT is the SUT. This couples BLT and ORCA a little more than I'd like, but I don't know a better solution. You can see this working here: https://github.com/acquia/blt/pull/4037

The only alternative I can think of is for ORCA to find whatever branch of BLT is checked out in Travis and pass that as the version parameter, and for BLT to configure the global path repo to point directly to the subtree split in Git rather than creating a tmp copy. But that seems difficult to set up and flaky.

### DIFF
--- a/config/packages.yml
+++ b/config/packages.yml
@@ -58,6 +58,9 @@ drupal/acsf: []
 acquia/blt:
   type: composer-plugin
   core_matrix:
+    '>=9.0.0':
+      version: 12.x
+      version_dev: 12.x-dev
     '>=8.8.0':
       version: 11.x
       version_dev: 11.x-dev
--- a/src/Fixture/FixtureCreator.php
+++ b/src/Fixture/FixtureCreator.php
@@ -346,6 +346,9 @@ class FixtureCreator {
     $version = ($this->isDev)
       ? $this->blt->getVersionDev($this->drupalCoreVersion)
       : $this->blt->getVersionRecommended($this->drupalCoreVersion);
+    if ($this->sut && $this->sut->getPackageName() === 'acquia/blt') {
+      $version = 'dev-master';
+    }
     $command = [
       'composer',
       'create-project',


TravisCarden:
@danepowell, in the absence of an explicit value in its `composer.json`, Composer determines what version to assign to a path repository based on the checked out Git branch. That's why ORCA requires you to specify your working branch in your `.travis.yml`, so it can check out a branch by that name. If you do that in your case (i.e., check out a `12.x` branch in `/tmp/blt-project` after cloning it, that should work. Would you give that a try and let me know if it doesn't do the trick?

danepowell:
tl;dr: that doesn't work, but I think I'm going to solve it from the BLT side by just removing the BLT Project subtree split entirely, and managing that as a separate project. On the ORCA side, this would be solved by moving away from BLT Project.

I tried copying the blt-project subtree split into `/tmp/blt-project`, then doing a `git init` and `git checkout -b 12.x`. Composer still doesn't use that local version, it downloads a remote version of blt-project.

I suspect the reason is that the local version doesn't share any commits with the remote version, and for whatever reason in this scenario Composer prefers a remote version. In other words, when ORCA tells Composer to get 12.x, Composer can _see_ that the 12.x branch exists locally but it doesn't have any reason to believe it's newer than what's available remotely. As far as I can tell, there's nothing I can do from the BLT side to force Composer to prefer the local version. I can't even fake the commit history because I think Travis does a shallow clone.

The only solution I'm aware of is to use a unique branch name locally (in this case of this patch, master) so that Composer has nowhere else to turn :smile:

PR #66, opened 2020-03-06
Removed Drupal Console to fix compatibility with Drupal 9.

danepowell:
Drupal Console is not compatible with Drupal 9 due to conflicting Symfony versions: https://github.com/hechoendrupal/drupal-console/issues/4220

Builds break when ORCA tries to add it to test fixtures, as in https://github.com/acquia/blt/pull/4037


TravisCarden:
Thanks, @danepowell!

PR #67, opened 2020-03-09
Fixed missing dev packages in fixture on Drupal 9.

danepowell:
I suspect that if the core version is something like `9.0.0-alpha1`, casting to float makes it `0` or at least something unexpected, and this causes dev packages to be missing.

By the way, I found some other float casts here if you want to check if there are similar bugs: https://github.com/acquia/orca/blob/develop/src/Utility/DrupalCoreVersionFinder.php


danepowell:
Ah, this actually isn't a bug. As long as the core version starts with a number, it'll get cast to a float as expected. My problem was that the core version was a loose Composer constraint (`^9.0.0-alpha1`), which got cast to `0`. Using version_compare doesn't help with that, although I still wonder if it's more appropriate than casting to float.

PR #74, opened 2020-03-19
Fixed invalid composer.json in fixture.

danepowell:
When the fixture is initialized, composer.json is invalid (fails `composer validate`) because of the extra metadata that was added during creation with a corresponding update to the lock file.

This causes failures in downstream tests that rely on composer.json being valid (such as BLT's).

Edit: the change to ORCA's own composer.lock is unrelated, but also probably necessary :smile:

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -6082,5 +6082,6 @@
         "ext-json": "*",
         "ext-sqlite3": "*"
     },
-    "platform-dev": []
+    "platform-dev": [],
+    "plugin-api-version": "1.1.0"
 }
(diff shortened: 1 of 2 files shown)


TravisCarden:
Thanks, @danepowell! I picked the change to the fixture creator. I'm not sure what the change to ORCA's own `composer.lock` is. Do I need it if I'm not providing a Composer plugin? My `composer.json` validates without it.

danepowell:
It also validates for me, and I don't think there's a functional impact either way. But if you run `composer update --lock` (using Composer 1.10 or greater) you should see the change to composer.lock.

This is due to a change in Composer 1.10: https://github.com/composer/composer/blob/master/CHANGELOG.md#1100-2020-03-10

Sorry I shouldn't have mixed both changes in this PR.

TravisCarden:
Ah, cool. I've updated my Composer version and committed the change separately. Thanks.

PR #75, opened 2020-03-21
Fixes #73: Don't include SUT vendor directory in tests.

### LINKED ISSUE #73: qa:automated-tests including broken vendor test cases
When I run the following: `ORCA_FIXTURE_DIR=/home/dane/packages/orca-build /home/dane/packages/orca/bin/orca qa:automated-tests --sut=acquia/blt --sut-only`

I get the following error:
```
> '/home/dane/packages/orca-build/vendor/bin/phpunit' '--colors=always' '--debug' '--configuration=/home/dane/packages/orca-build/docroot/core/phpunit.xml.dist' '--exclude-group=orca_ignore' '/home/dane/packages/orca-build/vendor/acquia/blt'
PHP Fatal error:  Uncaught Error: Class 'PHPUnit_Framework_TestCase' not found in /home/dane/packages/blt/vendor/behat/mink-goutte-driver/tests/Custom/InstantiationTest.php:7
```

I think what's happening here is that ORCA is telling PHPUnit to run all tests in `/home/dane/packages/orca-build/vendor/acquia/blt`, which is going to pick up non-BLT tests in the BLT vendor directory, including ones like the mink-goutte-driver one above that's not compatible with PHPUnit 8.

It seems like either ORCA should be more specific (i.e. only pass the `tests` directory to PHPUnit:
`/home/dane/packages/orca-build/vendor/acquia/blt/tests`), or there needs to be a way to exclude that vendor directory from phpunit.

danepowell:
Fixes #73 

Instead of providing the SUT directory as an argument to the PHPUnit CLI, this moves the SUT directory to a testsuite definition in phpunit.xml.dist.

This is necessary in order to exclude the SUT vendor directory from the testsuite. As near as I can tell, this is the _only_ way to exclude a directory from PHPUnit tests.

For consistency's sake, you may wish to move other CLI parameters (such as groups to include/exclude) to phpunit.xml.dist. Let me know and I can update this PR.

Finally, excluding the vendor directory is a bit of a heuristic solution. I've confirmed it solves the bug for BLT and Lightning, but conceivably a package could have third-party tests it wants to exclude somewhere other than the vendor directory. I'm okay with this approach for now, but just a thought.


src/Task/TestFramework/PhpUnitTask.php
```diff
@@ -212,11 +238,11 @@ protected function runPhpUnit(): void {
         '--debug',
         "--configuration={$this->fixture->getPath('docroot/core/phpunit.xml.dist')}",
         '--exclude-group=orca_ignore',
+        '--testsuite=sut',
```
TravisCarden, line 241: Because this is added unconditionally, it will also limit integrated tests to the SUT, so non-SUT tests will never be run.

danepowell:
Per Slack discussion, renaming the testsuite to "ORCA" is adequate since there's no other way to exclude a directory except via testsuite definition in phpunit.xml.

danepowell:
@TravisCarden ready for review again. You can see from the last two commits I added a test that fails when the vendor directory is included, and then revised the code to exclude it.

PR #76, opened 2020-03-26
Files directory initialized outside docroot.

danepowell:
@TravisCarden you asked the other day where the "sites" directory in the fixture root came from. This is the answer :smile:


TravisCarden:
Ah, good find! Thanks, @danepowell.

PR #78, opened 2020-03-30
jakub-onderka packages are abandoned

danepowell:
Installing ORCA via Composer throws this warning:

> Package jakub-onderka/php-parallel-lint is abandoned, you should avoid using it. Use php-parallel-lint/php-parallel-lint instead.

It appears that (according to at least a couple users and a co-maintainer) jakub-onderka is no longer maintaining packages, and users should switch to these forks:
- https://github.com/JakubOnderka/PHP-Parallel-Lint/pull/156
- https://github.com/JakubOnderka/PHP-Console-Highlighter/pull/21

I've looked through both issue queues and have seen zero protest from the maintainer, so I'm assuming these packages truly are abandoned, unless the maintainer has just been on vacation for several months or something.


danepowell:
Unfortunately it appears we need to use a dev version for now because of recent upstream fixes to support running outside the project directory:
- https://github.com/php-parallel-lint/PHP-Parallel-Lint/pull/2
- https://github.com/php-parallel-lint/PHP-Parallel-Lint/commit/de81f0abc17444d3070e2c0623950ebf434f3361#diff-32f5d7dbf80a3aeafb22a44ccda60465

danepowell:
@TravisCarden this is ready for review

TravisCarden:
Excellent! Thanks, @danepowell!

PR #83, opened 2020-05-13
Clarify that feature branche will be renamed to ORCA_SUT_BRANCH

scor:


TravisCarden:
Ah, thanks, @scor. As it happens, I tried to achieve this a mere two hours ago with 739afba. What do you think? Does that sufficiently clarify the matter? I tried to avoid technical details in favor of functional outcomes. Would it have prevented your confusion, or does it need to more explicitly technical?

scor:
satisfied with df52cce 👍 thanks

PR #84, opened 2020-05-18
Remove unnecessary ignore rule.

danepowell:
Previously, Annotations deprecated the `registerFile()` method in 1.x but did not provide any replacement in the 1.x branch, which is obviously not good, and forced us to ignore this deprecation warning: https://github.com/acquia/orca/issues/27

Now, thanks to this upstream fix (https://github.com/doctrine/annotations/pull/271), projects can safely remove calls to deprecated `registerFile()` in annotations 1.10. The new fallback mechanism in Annotations means that it should "just work" like before without an explicit call to `registerFile()`. Thus we no longer need this ignore rule.


TravisCarden:
Thanks, @danepowell!

PR #85, opened 2020-05-18
ORCA-72: Use Acquia Recommended Project instead of BLT Project for D9/BLT12.

danepowell:
Changes in this PR:
- Allow SUT-less ORCA jobs to run on Travis CI (since the project templates need to run ORCA but aren't a SUT themselves or associated with any SUT)
- Use the new project templates for D9 builds instead of the old BLT Project template.

Once we stop supporting D8/BLT11 we can remove BLT references entirely from ORCA, won't that be nice :)

Not sure exactly when that will be, since BLT 11 is scheduled to be EOL 2020Q4 while D8 is EOL 2021Q4. We'll need to reconcile that, but that's an issue for BLT...

### DIFF
--- a/bin/travis/script.sh
+++ b/bin/travis/script.sh
@@ -31,7 +31,7 @@ case "$ORCA_JOB" in
   "INTEGRATED_DEV") eval "orca qa:automated-tests --sut=$ORCA_SUT_NAME $SUT_ONLY" ;;
   "CORE_NEXT") eval "orca qa:automated-tests --sut=$ORCA_SUT_NAME $SUT_ONLY" ;;
   "D9_READINESS") eval "orca qa:automated-tests --sut=$ORCA_SUT_NAME --sut-only" ;;
-  "CUSTOM") eval "orca qa:automated-tests --sut=$ORCA_SUT_NAME ${ORCA_CUSTOM_TESTS_RUN_ARGS:=}" ;;
+  "CUSTOM") eval "orca qa:automated-tests ${ORCA_CUSTOM_TESTS_RUN_ARGS:=}" ;;
 esac
 
 if [[ "$ORCA_ENABLE_NIGHTWATCH" = "TRUE" && "$ORCA_SUT_HAS_NIGHTWATCH_TESTS" && -d "$ORCA_YARN_DIR" ]]; then
--- a/src/Fixture/FixtureCreator.php
+++ b/src/Fixture/FixtureCreator.php
@@ -377,7 +377,7 @@ class FixtureCreator {
     }
     if (getenv("ORCA_SUT_NAME") === 'acquia/drupal-recommended-project') {
       // Have to use some unique branch name or Composer won't check out the path repo.
-      $version = getenv("ORCA_SUT_BRANCH");
+      $version = 'dev-' . getenv("ORCA_SUT_BRANCH");
     }
     $command = [
       'composer',
--- a/src/Fixture/FixtureCreator.php
+++ b/src/Fixture/FixtureCreator.php
@@ -375,6 +375,10 @@ class FixtureCreator {
       $project_template = 'acquia/drupal-recommended-project';
       $version = ($this->isDev) ? 'dev-master' : '^1';
     }
+    if ($this->sut->getPackageName() === 'acquia/drupal-recommended-project') {
+      // Have to use some unique branch name or Composer won't check out the path repo.
+      $version = 'dev-bologna';
+    }
     $command = [
       'composer',
       'create-project',
--- a/src/Fixture/FixtureCreator.php
+++ b/src/Fixture/FixtureCreator.php
@@ -252,7 +252,7 @@ class FixtureCreator {
    *   In case of errors.
    */
   public function create(): void {
-    $this->createBltProject();
+    $this->createProject();
     $this->fixtureConfigurator->ensureGitConfig();
     $this->configureBltProject();
     $this->fixDefaultDependencies();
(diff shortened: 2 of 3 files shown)


bin/travis/script.sh
```diff
@@ -31,7 +31,7 @@ case "$ORCA_JOB" in
   "INTEGRATED_DEV") eval "orca qa:automated-tests --sut=$ORCA_SUT_NAME $SUT_ONLY" ;;
   "CORE_NEXT") eval "orca qa:automated-tests --sut=$ORCA_SUT_NAME $SUT_ONLY" ;;
   "D9_READINESS") eval "orca qa:automated-tests --sut=$ORCA_SUT_NAME --sut-only" ;;
-  "CUSTOM") eval "orca qa:automated-tests --sut=$ORCA_SUT_NAME ${ORCA_CUSTOM_TESTS_RUN_ARGS:=}" ;;
```
danepowell, line 34: Removing the `--sut` parameter here and in install.sh is necessary to support custom jobs when no SUT is present (i.e. for acquia/drupal-recommended-project). Otherwise ORCA tries to require a SUT, and that fails when I'm trying to test a project template.

If you don't want to straight-up remove it, maybe you could make `--sut=$ORCA_SUT_NAME` a default value for `ORCA_CUSTOM_TESTS_RUN_ARGS` instead? It just needs to be something that a downstream project can override and remove completely if necessary. Or else maybe there could be a CUSTOM_NO_SUT job.

src/Fixture/FixtureCreator.php
```diff
@@ -358,25 +358,37 @@ public function setSymlinkAll(bool $symlink_all): void {
...
+    $project_template = 'acquia/blt-project';
+    // Handle D9 differently, since BLT Project no longer exists.
+    if (version_compare($this->getResolvedDrupalCoreVersion(), 9, '>=')) {
+      $project_template = 'acquia/drupal-recommended-project';
+      $version = ($this->isDev) ? 'dev-master' : '^1';
+    }
+    // Handle the very special case of ORCA running as part of acquia/drupal-recommended-project builds.
+    if (getenv("ORCA_SUT_NAME") === 'acquia/drupal-recommended-project') {
+      // Have to use some unique branch name or Composer won't check out the path repo.
+      $version = 'dev-' . getenv("ORCA_SUT_BRANCH");
```
danepowell, line 382: I know this is ugly as sin, but I don't know what else to do. The problem is that Composer will _never_ utilize the path repo if the branch name (dev-master in the case of drupal-recommended-project) exists upstream somewhere. I have no idea why. I suspect that the branch that ORCA creates in `before_install.sh` on Travis does not have the full Git history, so Composer doesn't see it as being related to the upstream version.

I'm also not sure why this works for SUTs but not for fixtures (BLT Project had the same problem). With BLT, ORCA runs `composer require acquia/blt:12.9999.9999.999-dev` when `ORCA_SUT_BRANCH` is `12.x`. How the heck does that work?

src/Fixture/FixtureCreator.php
```diff
@@ -358,25 +358,37 @@ public function setSymlinkAll(bool $symlink_all): void {
...
     }
+    // Handle D8 fixtures (default).
     $version = ($this->isDev)
       ? $this->blt->getVersionDev($this->drupalCoreVersion)
       : $this->blt->getVersionRecommended($this->drupalCoreVersion);
+    $project_template = 'acquia/blt-project';
+    // Handle D9 differently, since BLT Project no longer exists.
+    if (version_compare($this->getResolvedDrupalCoreVersion(), 9, '>=')) {
+      $project_template = 'acquia/drupal-recommended-project';
+      $version = ($this->isDev) ? 'dev-master' : '^1';
```
danepowell, line 377: If you want to not hardcode the project template name and branch versions, maybe they could exist in the `package.yml` config, along with some sort of parameter to indicate that they are fixture templates and _not_ full SUTs.

wimleers:
/me is eagerly watching this :eagle: :eye:

src/Fixture/FixtureCreator.php
gabesullice: I'm currently in the process of trying to use an alternative project template over in https://github.com/acquia/acquia_migrate/pull/188. We cannot use `acquia/drupal-recommended-project` on its own (for reason I'd be happy to explain if you must be convinced 😉). I would _love_ to have the ability to override this in a supported way. Right now, I'm doing the MVP/hackiest-thing-imaginable of [committing a `.patch` and applying that patch to ORCA in my `.travis.yml`](https://github.com/acquia/acquia_migrate/blob/deprojectify/.config/orca/override_orca_project_template.patch).

danepowell: From an engineering standpoint, I agree it would be elegant to allow the fixture to be selectable (even if just between `acquia/drupal-minimal-project` and `acquia/drupal-recommended-project`). But from a mission standpoint, I think this violates the first rule of ORCA, which is to replicate a single prescribed user journey and ensure all company packages work together. The more configurability ORCA adds, the less harmonized the testing process becomes.

The decision isn't really mine to make, but I think I'd prefer to fix whatever compatibility issues exist between `acquia/drupal-recommended-project` and acquia_migrate.

gabesullice: The thing is that the migrate product is itself a prescribed user journey. It begins in the cloud UI and ends in a working Drupal 9 site. Along the way, components of the migrate product (not necessarily the `acquia_migrate` module) _generate_ a project level `composer.json` which is derived from `drupal/drupal-minimal-project` and has a suite of additional packages pre-`require`'d that are based on each individual customer's existing Drupal 7 site.

The reason I would like to create a custom project template is to exercise the code that generates that `composer.json` and ensure it results in an installable site. FWIW, it was @TravisCarden that suggested I explore creating a project template and looking at BLT for an example of how to do so.

danepowell: Ah yeah, well if the user journey is intentionally bifurcating and the only difference is the starting point (fixture base / project template) then yeah I guess ORCA needs to support that. Maybe you could still use acquia/drupal-minimal-project as a base.

danepowell:
@TravisCarden this is ready for your review. I'm fine with merging it as is since I've tested it in combination with https://github.com/acquia/drupal-recommended-project/pull/1 to ensure it works. But you may want to weigh in on larger architectural issues we've called out in above comments.

wimleers:
I just found out this is now the top priority for @TravisCarden, can't wait to see this happen!

src/Fixture/FixtureCreator.php
danepowell: Per Travis on Zoom, this version string is actually pulled from Composer's guts by ORCA.

TravisCarden:
Thanks, everyone! Fixed by https://github.com/acquia/orca/pull/89.

PR #88, opened 2020-05-20
See what happens if we remove mglaman package

danepowell:


danepowell:
As expected, this deconflicts other dependencies and allows `mglaman/drupal-check:1.1.1` to be installed.

PR #89, opened 2020-06-02
Add the ability to specify an arbitrary Composer project template

TravisCarden:
This PR adds the ability to specify any Composer project template for the test fixture through the use of a new `$ORCA_FIXTURE_PROJECT_TEMPLATE` environment variable which may be specified in your `.travis.yml`.

It also changes the default project template for the `D9_READINESS` job to [`acquia/drupal-recommended-project`](https://github.com/acquia/drupal-recommended-project), which will become Acquia's officially recommended starting point beginning with Drupal 9, when BLT will drop support for `acquia/blt-project`. Manual testing indicates that the change should have no material impact on out-of-the-box jobs.

With the possible exception of the default `D9_READINESS` project template change, no BC-breaking changes are introduced herein. The new feature is entirely opt-in. On the outside chance that it _does_ break your builds, just set the `ORCA_FIXTURE_PROJECT_TEMPLATE` variable to `acquia/blt-project` in your `.travis.yml` to override the change.


wimleers:
🥳

PR #91, opened 2020-06-11
Make Composer use the local repository of a project template SUT

### LINKED ISSUE #90: Project templates as SUTs
Thanks to https://github.com/acquia/orca/pull/89 , we can now run ORCA on arbitrary project templates / fixtures, which is awesome. 🎉 

However, I'm not clear on how to run ORCA tests from the templates themselves, i.e. with the template as a SUT. For instance, I'd like to run ORCA on the Acquia Recommended Project template, but as you can see, ORCA complains that `acquia/drupal-recommended-project` is not a valid SUT: https://github.com/acquia/drupal-recommended-project/pull/2

Looking through https://github.com/acquia/orca/pull/89 and seeing functions like `isProjectTemplateSut()`, it seems the intent is the project template _can_ itself be a SUT. That's fine by me, but if that's the intent, how shall I work around that error?

danepowell:
This makes progress on #90 by fixing a couple of minor issues with using project templates as SUTs, plus two major hacks that need to get ironed out:
- ORCA doesn't work its version magic when composer requiring the project template, so Composer never uses the path repo, hence the `dev-bologna` branch hack
- ORCA tries to run Drupal core tests, incorrectly interpreting them as the project-template's tests

### DIFF
--- a/src/Fixture/FixtureCreator.php
+++ b/src/Fixture/FixtureCreator.php
@@ -194,47 +194,7 @@ class FixtureCreator {
    */
   private function createComposerProject(): void {
     $this->output->section('Creating Composer project');
-    $this->codebaseCreator->create($this->options, $this->getProjectTemplateString());
-  }
-
-  /**
-   * Gets the project template package/constraint string.
-   *
-   * @return string
-   *   The project template package/constraint string, e.g.,
-   *   acquia/drupal-recommended-project or acquia/blt-project:12.x.
-   */
-  private function getProjectTemplateString(): string {
-    $project_template = $this->options->getProjectTemplate();
-    switch (TRUE) {
-      case $this->isSutProjectTemplate():
-        return "{$project_template}:dev-bologna";
-
-      case $project_template === 'acquia/blt-project':
-        $version = ($this->options->isDev())
-          ? $this->blt->getVersionDev($this->options->getCore())
-          : $this->blt->getVersionRecommended($this->options->getCore());
-        return "{$project_template}:{$version}";
-
-      default:
-        return $project_template;
-    }
-  }
-
-  /**
-   * Determines whether or not the Composer project template is also the SUT.
-   *
-   * @return bool
-   *   Returns TRUE if the Composer project template is also the system under
-   *   test or FALSE if not.
-   */
-  private function isSutProjectTemplate(): bool {
-    if (!$this->options->hasSut()) {
-      return FALSE;
-    }
-
-    /* @noinspection NullPointerExceptionInspection */
-    return $this->options->getSut()->isProjectTemplate();
+    $this->codebaseCreator->create($this->options);
   }
 
   /**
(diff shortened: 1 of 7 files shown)


src/Fixture/CompanyExtensionEnabler.php
```diff
@@ -187,7 +187,7 @@ private function getCompanyExtensionList(string $extension_type): array {
     $extension_list = [];
 
     $top_level_packages = $this->packageManager->getAll();
-    if ($this->isSutOnly) {
+    if ($this->isSutOnly && $this->sut->getPackageName() != 'acquia/drupal-recommended-project') {
```
danepowell, line 190: Obviously a hack, I don't know how to tell in this context whether the SUT is a project template

src/Fixture/FixtureCreator.php
```diff
@@ -409,7 +409,7 @@ private function createComposerProject(): void {
   private function getProjectTemplateString(): string {
     switch (TRUE) {
       case $this->isProjectTemplateSut():
-        return "{$this->projectTemplate}:@dev";
+        return "{$this->projectTemplate}:dev-bologna";
```
danepowell, line 412: This is necessary to convince Composer to use the path repo instead of remote package version. It seems like ORCA doesn't work its sut-branch-to-version-number magic on the project template like it does on other SUTs (i.e. requiring version 12.9999.9999.999 or whatever for BLT). If it did that, maybe we wouldn't need this hack.

src/Fixture/Package.php
```diff
@@ -195,6 +195,10 @@ public function getInstallPathRelative(): string {
       return $this->data['install_path'];
     }
 
+    if ($this->getPackageName() == 'acquia/drupal-recommended-project') {
```
danepowell, line 198: Hack. As you can see below, I tried to instead key this off of a custom "project-template" Composer project type, but that didn't work. Maybe you can't make up project types willy-nilly.

The fixture creator already knows if the Sut is a project template via `isProjectTemplateSut()`, can we expose that so it's useful here? Or could that actually be a property of the package/sut somehow?

src/Fixture/CompanyExtensionEnabler.php
```diff
@@ -140,8 +140,8 @@ private function enableAcquiaExtensions(): void {
     }
 
     $this->output->section('Enabling company modules & themes');
-    $this->enableModules();
-    $this->enableThemes();
+    //$this->enableModules();
```
danepowell, line 143: Hack. I don't understand why, but ORCA wants to enable my SUT as a Drupal module. Why on earth would ORCA think it's a Drupal module?

src/Task/TestFramework/TestRunner.php
```diff
@@ -127,7 +127,7 @@ public function run(): void {
       $this->startServers();
     }
 
-    if ($this->sut) {
+    if ($this->sut && $this->sut->getPackageName() != 'acquia/drupal-recommended-project') {
```
danepowell, line 130: ORCA interprets the Drupal core tests as the _template's_ tests, which is bad because core tests take hours to run.

danepowell:
@TravisCarden how would you like to proceed with this? I've outlined in the issue description the problems I'm trying to solve, and noted the fixes/hacks in comments.

This is somewhat urgent, because until we get it sorted I cannot run public / integrated ORCA tests when making changes to project templates, which risks regressions.

TravisCarden:
Thanks, @danepowell. I'm starting on this now. Per [the precedent set by Drupal core's project templates](https://www.drupal.org/docs/develop/using-composer/starting-a-site-using-drupal-composer-project-templates) (not to mention the need to support them and the fact that it seems most "correct" in principle), I'm going to treat `type: "project"` in `composer.json` as a project template. Obviously, you'll need to modify your templates accordingly.

danepowell:
@TravisCarden tests are actually passing on this. It's still hacky in parts, but the changes only affect Drupal Recommended Project so it should be safe to merge. This would avoid having to constantly rebase to track `develop`.

PR #92, opened 2020-06-16
Add support for Drupal Test Traits' Selenium2 driver

phenaproxima:
Right now, Drupal Test Traits supports two Mink drivers: a more esoteric JSON Wire one (https://gitlab.com/DMore/chrome-mink-driver) and the standard behat/mink-selenium2-driver. However, ORCA inadvertently only allows you to use the former, because it doesn't set the DTT_MINK_DRIVER_ARGS environment variable correctly, which is needed for the Selenium2 driver to work.

This PR adds support for this environment variable. I also took the liberty of doing some refactoring in order to reduce repetition which cropped up while implementing this.

### DIFF
--- a/src/Task/TestFramework/PhpUnitTask.php
+++ b/src/Task/TestFramework/PhpUnitTask.php
@@ -124,16 +124,9 @@ class PhpUnitTask extends TestFrameworkBase {
       ->item(0)
       ->setAttribute('bootstrap', "{$this->projectDir}/vendor/weitzman/drupal-test-traits/src/bootstrap.php");
 
-    if (!$xpath->query('//phpunit/php/env[@name="DTT_BASE_URL"]')->length) {
-      $element = $doc->createElement('env');
-      $element->setAttribute('name', 'DTT_BASE_URL');
-      $element->setAttribute('value', sprintf('http://%s', Fixture::WEB_ADDRESS));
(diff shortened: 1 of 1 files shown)


phenaproxima:
FYI, this may be a blocker to implementing some test coverage (specifically, Cohesion-related test coverage) in Acquia CMS.

src/Task/TestFramework/PhpUnitTask.php
```diff
@@ -44,16 +70,26 @@ public function execute(): void {
...
+    // the JSON string to &quot;, since the XML attribute value is itself
+    // enclosed in double quotes. There's no way to change this behavior, so we
+    // must do a string replacement in order to wrap the Mink driver arguments
+    // in single quotes.
+    // @see https://stackoverflow.com/questions/5473520/php-dom-and-single-quotes#5473718
+    $mink_arguments = $this->getMinkWebDriverArguments();
+    $search = sprintf('value="%s"', htmlentities($mink_arguments));
+    $replace = sprintf("value='%s'", $mink_arguments);
+    $xml = str_replace($search, $replace, $this->doc->saveXML());
+    file_put_contents($path, $xml);
```
TravisCarden, line 92: Let's move this into `::setMinkDriverArguments()`.

PR #104, opened 2020-08-24
Debugging info for failed SUT placement

danepowell:
I got a little confused trying to integrate ORCA with Drupal Environment Detector (https://github.com/acquia/drupal-environment-detector/pull/9) because ORCA assumed it was a Drupal module and threw an exception "Failed to place SUT at correct path", which by itself isn't very helpful in diagnosing the issue.

It turned out to be pretty simple, I just need to change the SUT package type, but the exception could do a better job of leading people to that info.

See an example of the output of this patch: https://travis-ci.com/github/acquia/drupal-environment-detector/jobs/377032879#L2373

### DIFF
--- a/src/Fixture/FixtureCreator.php
+++ b/src/Fixture/FixtureCreator.php
@@ -728,6 +728,7 @@ class FixtureCreator {
     $sut_install_path = $this->sut->getInstallPathAbsolute();
 
     if (!file_exists($sut_install_path)) {
+      $this->displayFailedPlaceDebuggingInfo();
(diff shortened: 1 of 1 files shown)


TravisCarden:
Thanks, @danepowell. I'm gonna chew on this a little bit. I'm wondering if more of this could be validated _before_ creating the fixture. 🤔

danepowell:
Not that this would have changed the outcome at all, but just for posterity, the correct package type was `library`: https://github.com/acquia/orca/pull/132

PR #107, opened 2020-09-15
Replace Amplitude with Domo for telemetry

japerry:
Should work, however the DOMO USER key and password are required to be put into Travis before we can test appropriately.


src/Console/Command/Internal/InternalCreateDataSetCommand.php
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function execute(InputInterface $input, OutputInterface $output): int {
+    $simulate = $input->getOption('simulate');
+
+    // Get the Telemetry Config
+    $data = $this->telemetryClient->getLocalConfig();
```
TravisCarden, line 60: Command classes handle CLI interaction only (i.e., user input and exit codes). Detailed logic is delegated to domain objects, like `\Acquia\Orca\Domain\Fixture\FixtureCreator` or `\Acquia\Orca\Domain\Fixture\FixtureInspector`. That makes testing both a lot easier. (Existing command classes currently have almost 100% test coverage.) Input that requires validation is done via options classes like `\Acquia\Orca\Domain\Fixture\FixtureOptions`. (Those currently have _exactly_ 100% test coverage.)

src/Console/Command/Internal/InternalCreateDataSetCommand.php
```diff
@@ -0,0 +1,91 @@
...
+    parent::__construct(self::$defaultName);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function configure() {
+    $this
+      ->setDescription('Creates/updates a domo data stream based on current telemetry config yml')
+      ->addOption('simulate', ['s'], InputOption::VALUE_NONE, 'Run in simulated mode: show what would be created instead of actually creating the stream')
```
TravisCarden, line 49: Nice. 👍

src/Helper/Log/TelemetryClient.php
```diff
@@ -44,26 +67,141 @@ public function __construct(Amplitude $amplitude, bool $telemetry_is_enabled, ?s
...
+    return (bool) $this->domoClient;
+  }
+
+  /**
+   * @param $data
+   * Dataset in JSON format
+   *
+   * @return object|array The result with the dataset ID, or an error.
+   */
+  public function createDataSet(array $data) {
```
TravisCarden, line 79: Per the Single Responsibility Principle, this class should have only one reason to change. As the Domo client, that reason should be due to API differences with Domo. Generating the data to send should belong to a class with only _that_ responsibility. See `\Acquia\Orca\Helper\Log\TelemetryEventPropertiesBuilder`, for example. In fact, you should modify that class according to your needs, renaming it to `OrcaEventPropertiesBuilder` if the logic is Domo-specific, and use it.

TravisCarden (changes requested):
Thanks for this PR, @japerry! As this is your first major PR, I've given a lot of feedback on it. That's because ORCA enjoys a great deal of internal consistency right now that I don't want to degrade, and because this is a good chance to "show you around the system". :)

A few high level things: Please...

- [ ] Add Domo to the project glossary (`docs/glossary.md`).
- [ ] Grep the codebase for "Amplitude" (incl. "amplitude"). I commented on a number of left behind instances I noticed in the PR, but not all of them, and I'm pretty sure there are others lying around.
- [ ] Make the tests pass.

Notes:

- When you rebase/resubmit this you're going run into some significant directory restructuring I recently performed. Sorry!
- Since this is an open-source repo, I don't refer to internal Jira tickets in it.

composer.json
```diff
@@ -79,6 +79,12 @@
             "Acquia\\Orca\\Tests\\": "tests/"
         }
     },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/eporama/domo-php"
```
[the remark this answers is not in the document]
japerry: Yes, the packagist version is pegged to an old version of guzzle. While there is a PR, it hasn't been acted upon in some time. Our own Erik Peterson forked it so I'm using his version instead. (I presume they use it for other Acquia projects as well)

japerry:
Thanks for the comments! Doing some refactoring as suggested. I also need to look at rewriting the tests for the amplitude related code. Hopefully in the next PR we can somehow test the username/id so I can see if its transmitting properly. But this is starting to take shape!

TravisCarden:
Closing as outdated.

PR #109, opened 2020-10-06
Update to PHP 7.3

danepowell:
This should probably be done in sync with https://github.com/acquia/coding-standards-php/pull/15

We explicitly base our PHP version on that recommended by Acquia and Drupal (see links in example/.travis.yml).

Drupal now recommends PHP 7.3, and Acquia has removed PHP 7.2 from the Cloud platform. PHP 7.2 itself is EOL in one month.

Let's go ahead and make PHP 7.3 the standard for ORCA.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         }
     ],
     "require": {
-        "php": ">=7.2",
+        "php": ">=7.3",
         "ext-dom": "*",
         "ext-json": "*",
         "ext-sqlite3": "*",
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         }
     ],
     "require": {
-        "php": ">=7.3",
+        "php": ">=7.2",
         "ext-dom": "*",
         "ext-json": "*",
         "ext-sqlite3": "*",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "ece3ff3c720b7aad4964cb08b30b4f9f",
+    "content-hash": "8cd8c336c15084c305a0e099cec82cfb",
     "packages": [
         {
             "name": "acquia/coding-standards",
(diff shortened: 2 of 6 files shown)


composer.json
```diff
@@ -21,7 +21,7 @@
         }
     ],
     "require": {
-        "php": ">=7.2",
```
TravisCarden, line 24: We'll need to leave this at 7.2 because [D8 still supports it](https://www.drupal.org/docs/system-requirements/php-requirements#php_required) and will indefinitely, i.e., 7.3 isn't the minimum required version until D9. Thus we'll default to 7.3 in our `.travis.yml` files, but ORCA needs to at least be installable on 7.2 until D8 is no longer supported.

docs/getting-started.md
```diff
@@ -31,7 +31,7 @@ See also [Continuous integration](understanding-orca.md#continuous-integration).
 
 ORCA can also be installed and run locally for testing and development. Follow these steps to set it up:
 
-1. Ensure you have PHP 7.2 or later with at least 256 MB of memory allocated to it and [Composer](https://getcomposer.org) installed.
+1. Ensure you have PHP 7.3 or later with at least 256 MB of memory allocated to it and [Composer](https://getcomposer.org) installed.
```
TravisCarden, line 34:
```suggestion
```
As before, 7.2 needs to be supported as long as D8 is.

composer.lock
```diff
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "d91b6375bf70c3bb6848882f9867eb94",
```
TravisCarden, line 7: Obviously, this file needs to be restored to match the corresponding reversion of `composer.json`.

TravisCarden (changes requested):
Thanks, @danepowell. A couple of changes needed. Note to self: we should document minimum requirements sometime. (That's not requested for this PR.)

composer.json
danepowell: Even if Drupal supports an end-of-life PHP version, I'm not sure that Acquia should. We've already removed it from our hosting platform. But it doesn't matter much to me, I'll change it back.

TravisCarden (approved):

PR #116, opened 2020-11-18
Add support for PHPUnit 9 in Drupal 9.1

TravisCarden:
Drupal core updated to PHPUnit 9 in version 9.1 ([change record](https://www.drupal.org/node/3176567)), causing some ORCA CI jobs to fail. This fixes them by requiring [`phpspec/prophecy-phpunit`](https://packagist.org/packages/phpspec/prophecy-phpunit).


danepowell:
This solves some compatibility issues but not all. See the failing tests in https://github.com/acquia/blt/pull/4285

In that case, BLT uses `assertFileNotExists` from PHPUnit 8. This method was deprecated in PHPUnit 9. This means that our existing code fails on PHPUnit 9. But if we rename that method to avoid the deprecation warning, PHPUnit 8 tests fail because the new method doesn't exist.

PR #118, opened 2020-11-24
DIT-161: Composer 2 support.

japerry:
Introduces more complete composer2 support, see if this passes ORCA tests.

Major updates:
 * Viamo chrome driver deprecated in place of Lanfest package
 * Dealerdirect updated to 0.7.0
 * taptima/phalyfusion forked to local repo for now, waiting for PR to be merged in master, but works with the repository patch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,6 @@ jobs:
 
 before_install:
   - nvm use 12.13.1
-  - composer self-update 1.10.16
   - ../orca/bin/travis/self-test/before_install.sh
   - ../orca/bin/travis/before_install.sh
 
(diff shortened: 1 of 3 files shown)


japerry:
Currently failing due to services.yml adding a $minimumStability flag to the dependency pools. This has be rewritten (so it seems) in composer 2, so the API is different.
However, lacking a change record, and not being familiar with this part of the composer subsystem, it didn't look obvious to me where you would be placing the minversion flags to get the desired result ORCA is looking for.

Hopefully @TravisCarden can help here.

danepowell:
This commit describes what changed in Composer, specifically I think ORCA needs to use the new pool builder: https://github.com/composer/composer/commit/c0f19f6c573550916373fbaae7245558ffb75af6

TravisCarden:
Fixed in https://github.com/acquia/orca/releases/tag/v3.5.0. Thanks, @japerry.

PR #123, opened 2020-12-23
Update phpmd to 2.9.1

kaynenh:
Content Hub's most recent static analysis improvements are blocked on an error (https://travis-ci.com/github/acquia/acquia_contenthub/jobs/464126513) caused by phpmd failing to properly handle reflection classes (as best as I can tell). According to this, it should be fixed in 2.9.1 - https://github.com/phpmd/phpmd/issues/816#issuecomment-687215399

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
         "php-parallel-lint/php-console-highlighter": "^0.4.0",
         "php-parallel-lint/php-parallel-lint": "^1.2.0",
         "phploc/phploc": "^5.0",
-        "phpmd/phpmd": "^2.6",
+        "phpmd/phpmd": "^2.9.1",
         "stecman/symfony-console-completion": "^0.10.1",
         "symfony/config": "^4.1",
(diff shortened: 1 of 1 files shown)


kaynenh:
Looked into this a bit more and realized the real issue here is that the lock file is set to 2.9.0 - https://github.com/acquia/orca/blob/64546197fc68359cf9eb450d6157b29aa6226ddf/composer.lock#L2698

Not sure this PR is exactly what's needed, but I can confirm 2.9.1 locally on my machine does not have this issue while it does happen in Travis on 2.9.0. If there's a way to force a rebuild of the lock file, that would work too.

TravisCarden:
This has been fixed in #124, @kaynenh. Obviously, it won't be released until the New Year. If you want to use it in the meantime, change `$ORCA_VERSION` to `dev-develop` in your `.travis.yml`, e.g.,

```diff
-    - ORCA_VERSION=^3
+    - ORCA_VERSION=dev-develop
```

PR #130, opened 2021-02-08
Fix command not found error in before_cache.sh

danepowell:
This script produces a harmless but potentially confusing error: https://travis-ci.com/github/acquia/automatic-updater/jobs/481387635#L2229

> ../orca/bin/travis/before_cache.sh: line 13: orca: command not found


TravisCarden:
Ooh thanks, @danepowell. That may be harmless now, but it's definitely a defect. Merging.

PR #131, opened 2021-02-09
Make sed command in example .travis.yml more cross-platform portable

danepowell:
The provided sed command works on Mac but fails on Unix (e.g. Ubuntu).

Mac uses BSD sed, which has slightly different syntax than Unix. This is the most succinct cross-platform compatible command.


TravisCarden:
Excellent. Thanks, @danepowell! Merging.

PR #132, opened 2021-02-09
Add acquia/drupal-environment-detector to packages

danepowell:
The environment detector has been integrated with ORCA for some time, it would be nice to include it in integrated tests and not maintain a packages_alter: https://github.com/acquia/drupal-environment-detector/blob/master/tests/packages_alter.yml

### DIFF
--- a/config/packages.yml
+++ b/config/packages.yml
@@ -104,7 +104,7 @@ acquia/coding-standards:
   url: ../coding-standards-php
 
 acquia/drupal-environment-detector:
-  type: package
+  type: library
   version_dev: "dev-master"
 
 drupal/cog:
--- a/config/packages.yml
+++ b/config/packages.yml
@@ -103,6 +103,10 @@ acquia/coding-standards:
   type: phpcodesniffer-standard
   url: ../coding-standards-php
 
+acquia/drupal-environment-detector:
+  type: package
+  version_dev: "dev-master"
+
 drupal/cog:
   type: drupal-theme
   core_matrix:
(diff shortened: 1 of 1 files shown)


config/packages.yml
```diff
@@ -103,6 +103,10 @@ acquia/coding-standards:
   type: phpcodesniffer-standard
   url: ../coding-standards-php
 
+acquia/drupal-environment-detector:
+  type: package
```
TravisCarden, line 107: Let's change this to the more standard "library", [which is Composer's default](https://getcomposer.org/doc/04-schema.md#type):

> Out of the box, Composer supports four types:
>
> - **library:** This is the default. It will copy the files to `vendor`.

```suggestion
  type: library
```

TravisCarden (changes requested):
Ah, Im glad to see this. I just have one requested change.

config/packages.yml
danepowell: Thanks, I don't know why I chose `package` originally, maybe I didn't realize that it was based on Composer types 🤷 

I tested `library` in https://github.com/acquia/drupal-environment-detector/pull/15 and it seems to work.

config/packages.yml
```diff
@@ -103,6 +103,10 @@ acquia/coding-standards:
   type: phpcodesniffer-standard
   url: ../coding-standards-php
 
+acquia/drupal-environment-detector:
+  type: library
+  version_dev: "dev-master"
```
TravisCarden, line 108: You know what? Now that I think of it, `version_dev` isn't strictly necessary here.

TravisCarden (approved):

PR #139, opened 2021-02-23
Update versions of drupal/acquia_connector and drupal/acquia_search in packages.yml

japerry:
Acquia Search replaces Acquia Search Solr. It uses semantic versioning and only supports Drupal 8.9+
Connector 3.x also uses semantic versioning and only supports Drupal 8.9+

Hopefully we don't need to support Drupal 8.8?

### DIFF
--- a/config/packages.yml
+++ b/config/packages.yml
@@ -48,13 +48,8 @@
 # example.
 
 drupal/acquia_connector:
-  core_matrix:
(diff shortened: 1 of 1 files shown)


TravisCarden:
No, we don't need to support Drupal 8.8. Thanks, @japerry. Merging.

PR #151, opened 2021-08-27
Support Github Actions

danepowell:
Proof of concept: https://github.com/acquia/blt/pull/4393

- [x] Import template from BLT
- [x] Get ORCA CI itself on Actions

### DIFF
--- a/bin/travis/script.sh
+++ b/bin/travis/script.sh
@@ -21,6 +21,7 @@ if [[ "$ORCA_ENABLE_NIGHTWATCH" == "TRUE" && "$ORCA_SUT_HAS_NIGHTWATCH_TESTS" &&
     orca fixture:run-server &
     SERVER_PID=$!
 
+    # @todo could we set DRUPAL_TEST_CHROMEDRIVER_AUTOSTART instead of launching Chromedriver manually?
     $ORCA_ROOT/vendor/bin/chromedriver --disable-dev-shm-usage --disable-extensions --disable-gpu --headless --no-sandbox --port=4444 &
     CHROMEDRIVER_PID=$!
 
--- a/bin/travis/script.sh
+++ b/bin/travis/script.sh
@@ -19,15 +19,17 @@ if [[ "$ORCA_ENABLE_NIGHTWATCH" == "TRUE" && "$ORCA_SUT_HAS_NIGHTWATCH_TESTS" &&
   (
     cd "$ORCA_YARN_DIR" || exit
     orca fixture:run-server &
-    PID=$!
+    SERVER_PID=$!
 
     $ORCA_ROOT/vendor/bin/chromedriver --disable-dev-shm-usage --disable-extensions --disable-gpu --headless --no-sandbox --port=4444 &
+    CHROMEDRIVER_PID=$!
 
     eval "yarn test:nightwatch \\
       --headless \\
       --passWithNoTests \\
       --tag=$ORCA_SUT_MACHINE_NAME"
 
-    kill -0 $PID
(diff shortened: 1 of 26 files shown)


danepowell:
Nevermind the above comment about failing tests: I was able to downgrade Xdebug and change TTY detection in order to fix everything.

danepowell:
This is so close, but something screwy is still happening with TTY detection where I think ORCA is waiting for input and causing jobs to stall in Github Actions. See the latest failing test in https://github.com/acquia/blt/pull/4393

Funnily, it works on Travis CI (in this PR).

I'm out the next few days if anyone wants to take a peak in my absence and help get this over the finish line.

wimleers:
Exciting! 🤓 🥳

danepowell:
Still stumped on this. I've pared down BLT to a single simple unit test that definitely doesn't use STDIN, so I don't think it's the test case itself causing the job to stall. And yet, non-phpunit jobs run to completion, so it does seem to be specific to phpunit or the `qa:automated-tests` command. My hunch is that the process holds on to some other file / process handle and so Symfony never releases the process.

Could try with `no-servers` flag I guess.

danepowell:
Looks like my hunch was right... although adding the flag required reworking the commands a bit. But something about the latest commits fixed this.

danepowell:
Interesting, servers have nothing to do with it. My next wrong theory is that this is due to the redispatching between `orca ci:run` and `orca qa:automated-tests"

danepowell:
Travis and I determined that the tests stalling out was caused by using `Process::start()` on a process that we backgrounded (i.e. `chromedriver &`. Furthermore, this was only a problem when ORCA redispatched from `ci:run` to `qa:automated-tests`, it wasn't a problem when running `qa:automated-tests` directly. It's not totally clear why this is, but at some point we must have lost track of or failed to kill one of these processes, and that caused Symfony to wait forever to finish the parent process.

Since we were already using `Process::start()`, explicitly backgrounding the process with `&` should be redundant, and indeed bringing the process to the foreground seemed to fix everything.

Next time this happens, maybe we'd have more luck troubleshooting by setting a short timeout, and then inspecting running processes when the job fails to see if one has been orphaned. Symfony should really have an error handler (if it doesn't already) that explicitly says "died waiting on process 'chromedriver' to finish" or something.

danepowell:
Yarrrggghhh...

@TravisCarden see the latest test failures. I suspect Chromedriver is not in fact running in the background following our latest changes 😢

bin/travis/script.sh
```diff
@@ -19,13 +19,17 @@ if [[ "$ORCA_ENABLE_NIGHTWATCH" == "TRUE" && "$ORCA_SUT_HAS_NIGHTWATCH_TESTS" &&
   (
     cd "$ORCA_YARN_DIR" || exit
     orca fixture:run-server &
-    PID=$!
+    SERVER_PID=$!
+
+    $ORCA_ROOT/vendor/bin/chromedriver --disable-dev-shm-usage --disable-extensions --disable-gpu --headless --no-sandbox --port=4444 &
```
danepowell, line 24: There's a bug in the current implementation of `orca qa:automated-tests`: it's supposed to kill Chromedriver on exit, but instead leaves it running. Nightwatch actually depended on this bug in order to run its own tests.

Since we've fixed that in this PR (Chromedriver now exits cleanly along with ORCA), Nightwatch needs to start its own server.

danepowell:
Now that this is basically working, I'm going to tidy it up in https://github.com/acquia/orca/pull/152

PR #153, opened 2021-09-21
Stop on errors in _includes.sh

danepowell:
Katherine and I completely missed a syntax error while working _includes.sh, because Bash doesn't exit immediately upon errors.

@TravisCarden do you recall why you only `set -e` at the _end_ of _includes.sh, rather than the beginning?

### DIFF
--- a/bin/ci/_includes.sh
+++ b/bin/ci/_includes.sh
@@ -9,9 +9,8 @@
 # DESCRIPTION
 #     Includes common features used by ORCA scripts.
 
-# Exit as soon as one command returns a non-zero exit code and make the shell
-# print all lines in the script before executing them.
-set -ev
+# Exit as soon as one command returns a non-zero exit code.
+set -e
 
 # Outputs a formatted error message and exits with an error code if a given
 # condition is not met.
--- a/bin/ci/_includes.sh
+++ b/bin/ci/_includes.sh
@@ -93,3 +93,6 @@ export PATH="$HOME/.phpenv/shims/:$PATH"
 
 # Add convenient aliases.
 alias drush='drush -r "$ORCA_FIXTURE_DIR"'
+
+# Make the shell print all lines in the script before executing them.
++ set -v
--- a/bin/ci/_includes.sh
+++ b/bin/ci/_includes.sh
@@ -9,6 +9,10 @@
 # DESCRIPTION
 #     Includes common features used by ORCA scripts.
 
+# Exit as soon as one command returns a non-zero exit code and make the shell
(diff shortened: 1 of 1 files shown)


danepowell:
The failing test isn't related to this PR.

I think we need to address the failing test. It's "allowed" to fail and won't block merging this PR, but will inevitably cause confusion (I hate the red 'x').

It's failing because of the ACMS transition. I pinged Travis with details in Slack. We'd either need to fix the test so it passes (i.e. stop scanning failing contrib modules), or disable it.

danepowell:
Regarding the test failures, we're going to try to fix the failing tests. For the live job, I've opened https://github.com/acquia/orca/pull/154 . For the deprecated code scan, we decided isolated tests should only include the SUT as a dependency, and therefore should use the minimal project template. Since the minimal project template contains no failing code, this should pass again.

SUTs that do have deprecated code will see failures in their own isolated tests. This is perhaps an improvement over Travis CI, which swept these failures under the rug by not marking the whole job as failed. Note that this is a "soft" failure, it won't block PR merges.

bin/ci/_includes.sh
```diff
@@ -9,6 +9,10 @@
 # DESCRIPTION
 #     Includes common features used by ORCA scripts.
 
+# Exit as soon as one command returns a non-zero exit code and make the shell
+# print all lines in the script before executing them.
+set -ev
```
TravisCarden, line 14:
```suggestion
# Exit as soon as one command returns a non-zero exit code.
set -e
```

bin/ci/_includes.sh
```diff
@@ -90,7 +94,3 @@ export PATH="$HOME/.phpenv/shims/:$PATH"
 
 # Add convenient aliases.
 alias drush='drush -r "$ORCA_FIXTURE_DIR"'
-
-# Exit as soon as one command returns a non-zero exit code and make the shell
-# print all lines in the script before executing them.
-set -ev
```
TravisCarden, line 96:
```diff
- # Exit as soon as one command returns a non-zero exit code and make the shell
- # print all lines in the script before executing them.
- set -ev
+ # Make the shell print all lines in the script before executing them.
+ set -v
```

TravisCarden (changes requested):
@danepowell, I moved the whole line (`set -ev`) to the bottom of `_includes.sh` so as not to needlessly print the whole contents of the file each of the (many) times it gets executed on every build. But looking at it again, there's no reason to move both switches together. Let's have our cake and eat it, too. 😉

TravisCarden (approved):

PR #157, opened 2021-09-28
Use drupal-minimal-project for isolated tests

danepowell:
Isolated jobs currently use acquia/drupal-recommended-project as a fixture. This is problematic because drupal-recommended-project requires ACMS, and therefore essentially every company package. This defeats the purpose of an isolated build, which is to catch any undeclared dependency (including Composer dependencies) for a SUT.

This is a regression caused by the switch from Lightning to Acquia CMS. With Lightning, isolated tests included minimal dependencies (despite using the same template) because ORCA explicitly removed Lightning from the fixture before re-adding the SUT.

Happily, this also fixes that last failing deprecated code scan job, so the build is green again. (It was scanning all of the contrib modules from drupal-recommend-project).

### DIFF
--- a/src/Options/FixtureOptions.php
+++ b/src/Options/FixtureOptions.php
@@ -296,14 +296,27 @@ class FixtureOptions {
    *   The package name.
    */
   public function getProjectTemplate(): string {
+    // Allow users to override default templates via option.
     if ($this->options['project-template']) {
       return $this->options['project-template'];
     }
+
+    // The new project templates only support D9+. Use BLT Project for D8.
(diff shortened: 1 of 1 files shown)


danepowell:
You can see that the minimal project is being used for isolated tests: https://github.com/acquia/orca/runs/3736424515#step:6:37
Meanwhile, recommended project still used for integrated tests: https://github.com/acquia/orca/runs/3736424674#step:6:36

This seems to shave ~45s off of isolated jobs.

TravisCarden:
Perfect; merging. We're racking up the PRs, @danepowell. Just let me know when you're ready for a release.

PR #159, opened 2021-09-30
Xdebug never disabled on Travis CI

danepowell:
Xdebug is no longer removed when `ORCA_COVERAGE_ENABLE=FALSE`. Xdebug _is_ removed if ORCA_COVERAGE_ENABLE is empty, but I don't think that's ever the case: _includes.sh sets it to either TRUE or FALSE.

Additionally, for builds on Xenial which uses Xdebug 3, this triggers a bug in phploc. This is a regression introduced by https://github.com/acquia/orca/pull/152

- BLT PR verifying that this patch works to disable Xdebug: https://github.com/acquia/blt/pull/4403
- Upstream phplocbug: https://github.com/sebastianbergmann/phploc/issues/222
- Follow-up issue to fix Xdebug 3 compatibility: https://github.com/acquia/orca/issues/160

@TravisCarden where would be the best place to add an automated test to ensure that Xdebug is disabled when `ORCA_COVERAGE_ENABLE=FALSE`?

### DIFF
--- a/bin/ci/before_install.sh
+++ b/bin/ci/before_install.sh
@@ -24,9 +24,18 @@ echo "$TRAVIS_NODE_VERSION"
 yarn --version
 
 # Disable Xdebug except on code coverage jobs.
-if [[ ! "$ORCA_COVERAGE_ENABLE" ]]; then
+if [[ ! "$ORCA_COVERAGE_ENABLE" == TRUE ]]; then
   if [[ "$TRAVIS" ]]; then phpenv config-rm xdebug.ini; fi
-  if [[ "$GITHUB_ACTIONS" ]]; then sudo phpdismod -v ALL xdebug; fi
+  if [[ "$GITHUB_ACTIONS" ]]; then
+    # phpdismod would be simpler but flaky
+    # @see https://github.com/shivammathur/setup-php/issues/350#issuecomment-735370872
+    scan_dir=$(php --ini | grep additional | sed -e "s|.*: s*||")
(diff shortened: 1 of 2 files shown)


TravisCarden:
This seems like the perfect place for an automated test to me, @danepowell: https://github.com/acquia/orca/blob/v3.10.0/bin/ci/self-test/script.sh#L26-L31. Something like this would do the trick, I think. What do you think?

```diff
+  XDEBUG_IS_ENABLED=$(php -r 'echo function_exists("xdebug_get_code_coverage") ? "TRUE" : "FALSE";')
 
   if [[ "$ORCA_COVERAGE_ENABLE" == TRUE ]]; then
+    if [[ "$XDEBUG_IS_ENABLED" == "FALSE" ]]; then
+      echo "ORCA_COVERAGE_ENABLE is on but Xdebug is disabled"
+      exit 1
+    fi
     eval './vendor/bin/phpunit --coverage-clover="$ORCA_SELF_TEST_COVERAGE_CLOVER"'
   else
+    if [[ "$XDEBUG_IS_ENABLED" == "FALSE" ]]; then
+      echo "ORCA_COVERAGE_ENABLE is off but Xdebug is enabled"
+      exit 1
+    fi
     eval './vendor/bin/phpunit'
   fi
```

wimleers:
Confirming. This caused https://github.com/acquia/acquia_migrate/pull/545.

TravisCarden (approved):

PR #161, opened 2021-09-30
ORCA-232: Test failures due to root cert expiration

danepowell:
More details upstream: https://github.com/composer/composer/issues/10121

I suspect that upgrading the CA bundle will fix this.


danepowell:
I'm fairly confident that this can't be worked around by the application. We'll need to go upstream.

PR #163, opened 2021-10-07
ORCA-197: PHP 8 compatibility

danepowell:
This PR upgrades ORCA's dependencies to support PHP 8. Specifically:
- Upgrade PHPUnit, phploc, php-console-highlighter, and binary-chromedriver to their PHP 8-compatible versions
- Upgrade to Xdebug 3 in GitHub Actions (Xdebug 2 does not support PHP 8)
- Convert all test classes to extend a base class with the ProphecyTrait (since the PHPUnit base TestCase class no longer includes the prophesize function)

To do:
- [x] Resolve PHPUnit upgrade impact on downstream projects
- [x] Fix PHPUnit deprecations
- [x] Fix webdriver-binary-downloader and binary-chromedriver: https://github.com/LANFest/webdriver-binary-downloader/pull/4
- [x] Remove `--ignore-platform-req=php` (including templates)
- [x] Upgrade GitHub Actions (including template) to Xdebug 3


danepowell:
@TravisCarden to support PHP 8 with ORCA, we'll need to update PHPUnit from 6 to 8 (or preferably 9). Is this going to require all ORCA consumers to update their own tests to support PHPUnit 8/9? I think the answer is yes, in which case is this still desirable? I imagine it requires a new major version of ORCA.

danepowell:
This is ready to go, but blocked on getting tests passing in #174

PR #173, opened 2021-11-01
ORCA-248: Remove allowed failures from GitHub Actions

danepowell:
Patch failures on next minor jobs are causing a widespread headache:
- https://github.com/acquia/blt/pull/4407
- https://github.com/acquia/drupal-environment-detector/pull/23
- https://github.com/acquia/drupal-recommended-project/pull/36

I think next minor jobs should be considered unstable and allowed to fail.

But stepping back, I think we should remove the concept of allowed failures altogether. It just doesn't work well with a binary pass/fail CI system (which is to say, pretty much _every_ CI system). Rather, I'd like to organize tests into those that _must_ run (i.e. integrated jobs), _should_ run (isolated jobs), and _can_ run (previously allowed failures).

Must-run jobs must run because a failure could cause other ORCA project builds to fail. Should-run jobs form the core of the ORCA test matrix, but failures are isolated to the current project so it's okay if these need to be modified or disabled temporarily per-project. 

Can-run jobs are risky enough that they aren't enabled by default. If individual projects want to enable a risky test for specific purposes (i.e. if ACMS wants to ensure compatibility with next_minor_dev), great! They can do so. But projects should enable these tests on their own schedule, and _then ensure that they continue to pass_. There should never be a concept of an "allowed failure".

Downstream projects wishing to incorporate these changes should remove allowed-failure jobs and INTEGRATED_TEST_ON_NEXT_MINOR from their builds (unless they have a reason to keep them), remove the allow-failure matrix parameter, and optionally reorganize the builds into "must run" and "should run" groups.


danepowell:
Per scrum today, we've decided to deal with the problem of allowed failures in a slightly different way. We're going to remove the concept of allowed failures altogether, since we're kind of abusing GitHub's `continue-on-error` function as it is. Instead, we'll leave those jobs in the build matrix but when there _is_ a failure due to upstream changes (as is the case currently with NEXT_MINOR jobs), we'll disable it via an ORCA hotfix release.

PR #174, opened 2021-11-04
Migrate "allowed failure" functionality to ORCA core

danepowell:
Currently we implement the concept of "allowed failures" at the level of individual CI systems. Travis CI handles this natively, while in GitHub Actions we've somewhat abused the "continue-on-error" method. Acquia CI does not handle it at all.

ORCA needs to handle allowed failures internally in order for them to be portable to all CI platforms. This has the nice side effect of allowing ORCA to temporarily allow jobs to fail (via a hotfix) in instances of an outage, so that individual teams don't need to modify their CI scripts.

This includes a guard condition to preserve backwards compatibility with Travis CI. This saves us from having to communicate the change or cut a new major release.

I considered handling this in the ORCA PHP command itself, but the problem with that approach is that we can never recover from PHP fatal errors. So the Bash scripts seem like the best place to handle it.


danepowell:
@TravisCarden do you have time to review this this week? It fixes all of the failing jobs on ORCA and ORCA-integrated projects.

TravisCarden (approved):

PR #175, opened 2021-11-05
ORCA-152: Audit packages.yml and D8 EOL

danepowell:
Auditing `packages.yml` and D8 EOL needs to be handled in a single PR because there's some code lurking in the fixture manager that's pretty tightly coupled to old BLT versions.

@TravisCarden would you prefer to manually specify a release branch for all packages in packages.yml? This is safest in terms of guaranteeing that fixtures are always built from the active release branch, but it's also work for maintainers to keep updated.


danepowell:
This is unfortunate... D8 is still marked as supported on d.o and in the [d.o release API](https://updates.drupal.org/release-history/drupal/current), despite being officially unsupported as of a few days ago. I think we'll have to wait for that to be fixed before ORCA stops trying to install D8.

config/packages.yml
```diff
@@ -86,19 +71,8 @@ drupal/acquia_search:
...
-      version: 11.x
-      version_dev: 11.x-dev
-    8.7.x:
-      version: 10.x
-      version_dev: 10.x-dev
-    8.6.x:
-      version: 9.x
-      version_dev: 9.x-dev
+  version: 13.x
+  version_dev: 13.x-dev
```
danepowell, line 75: I'm not sure if this will work because of a stale BLT branch alias that I can't get rid of without breaking BC. Need to look into it.

danepowell: I created a 12.x branch and updated the dev-main alias to be 13.x-dev, which I think should fix it...

danepowell:
This is ready to go since we hardcoded D8 as EOL in https://github.com/acquia/orca/pull/190

TravisCarden:
So much special case code removed! Merging.

PR #178, opened 2021-11-15
ORCA-228: Enable Coveralls for GitHub Actions

danepowell:
You can see this in action at https://github.com/acquia/blt/pull/4420

### DIFF
--- a/bin/ci/_includes.sh
+++ b/bin/ci/_includes.sh
@@ -78,9 +78,17 @@ export DRUPAL_TEST_WEBDRIVER_CHROME_ARGS="--disable-gpu --headless --no-sandbox"
 export DRUPAL_TEST_WEBDRIVER_HOSTNAME="localhost"
 export DRUPAL_TEST_WEBDRIVER_PORT="4444"
 
-# Set TMPDIR on GitHub Actions
-if [[ "$RUNNER_TEMP" ]]; then
-  export TMPDIR="$RUNNER_TEMP"
+if [[ ! "$ORCA_TEMP_DIR" ]]; then
+  # Travis CI.
+  if [[ "$TMPDIR" ]]; then
+    export ORCA_TEMP_DIR="$TMPDIR"
+  # GitHub Actions.
+  elif [[ "$RUNNER_TEMP" ]]; then
+    export ORCA_TEMP_DIR="$RUNNER_TEMP"
+  # Fallback default.
+  else
+    export ORCA_TEMP_DIR="/tmp"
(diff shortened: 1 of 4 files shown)


bin/ci/_includes.sh
```diff
@@ -78,6 +78,11 @@ export DRUPAL_TEST_WEBDRIVER_CHROME_ARGS="--disable-gpu --headless --no-sandbox"
 export DRUPAL_TEST_WEBDRIVER_HOSTNAME="localhost"
 export DRUPAL_TEST_WEBDRIVER_PORT="4444"
 
+# Set TMPDIR on GitHub Actions
+if [[ "$RUNNER_TEMP" ]]; then
```
danepowell, line 82: The Coveralls command in the after_success script relies on TMPDIR being set, but this was a TravisCI-ism. It's not set on GH Actions.

example/.github/workflows/orca.yml
```diff
@@ -79,8 +79,10 @@ jobs:
       # @see https://github.com/acquia/orca/blob/main/docs/advanced-usage.md#ORCA_AMPLITUDE_API_KEY
       # ORCA_TELEMETRY_ENABLE: "TRUE"
       #
-      # Do not modify the following line. It uses the build matrix to set ORCA_JOB for each job.
+      # Do not modify the following lines. They are critical for managing the build matrix.
       ORCA_JOB: ${{ matrix.orca-job }}
+      ORCA_COVERALLS_ENABLE: ${{ matrix.coveralls-enable }}
+      COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
TravisCarden, line 85:
```suggestion
      # The GITHUB_TOKEN secret is automatically available to GitHub Actions
      # workflows--it does not need to be manually created or configured anywhere.
      # @see https://docs.github.com/en/actions/security-guides/automatic-token-authentication
      COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

bin/ci/_includes.sh
TravisCarden: Great. However, before this change, the variable in that script (https://github.com/acquia/orca/blob/v3.11.0/bin/ci/after_success.sh#L25) was overridable--via bash's (ridiculous, convoluted [parameter expansion](https://unix.stackexchange.com/questions/122845/using-a-b-for-variable-assignment-in-scripts/122878) 🙄 ). Adding a commit to restore that functionality.

PR #179, opened 2021-11-15
Code coverage support for non-module packages

danepowell:
Code coverage reporting is currently broken for ORCA packages that are not Drupal modules (BLT, Drupal Environment Detector, etc), because Drupal core's phpunit.xml excludes non-modules from coverage reporting.

You can see this working at https://github.com/acquia/blt/pull/4420

### DIFF
--- a/src/Domain/Tool/Phpunit/PhpUnitTask.php
+++ b/src/Domain/Tool/Phpunit/PhpUnitTask.php
@@ -66,6 +66,7 @@ class PhpUnitTask extends TestFrameworkBase {
     $this->ensureSimpleTestDirectory();
     $this->setSimpletestSettings();
     $this->setTestSuite();
+    $this->setCoverageFilter();
     $this->enableDrupalTestTraits();
     $this->disableSymfonyDeprecationsHelper();
     $this->setMinkDriverArguments();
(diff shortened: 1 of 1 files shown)


src/Domain/Tool/Phpunit/PhpUnitTask.php
```diff
@@ -108,6 +109,26 @@ private function setTestSuite(): void {
...
+    $exclude = $this->doc->createElement('exclude');
+    $exclude_directory = $this->doc->createElement('directory', "{$this->getPath()}/tests");
+    $exclude->appendChild($exclude_directory);
+    $this->xpath->query('//phpunit/filter/whitelist')
+      ->item(0)
+      ->appendChild($directory);
+    $this->xpath->query('//phpunit/filter/whitelist')
+      ->item(0)
+      ->appendChild($exclude);
+  }
```
TravisCarden, line 130:
```suggestion
  /**
   * Sets code coverage filters to support non-Drupal module SUTs.
   *
   * Drupal core's phpunit.xml.dist sets a code coverage whitelist with test
   * files exclusion for Drupal modules, but it does not support themes,
   * profiles, or non-Drupal packages. Since ORCA must report test coverage for
   * all these, it must manually add the appropriate directories for the SUT.
   * The resulting additions look like the following:
   *
   * ```xml
   * <phpunit>
   *   <filter>
   *     <whitelist>
   *       <directory>/var/www/docroot/modules/contrib/example</directory>
   *       <exclude>
   *         <directory>/var/www/docroot/modules/contrib/example/tests</directory>
   *       </exclude>
   *     </whitelist>
   *   </filter>
   * </phpunit>
   * ```
   */
  private function setCoverageFilter(): void {
    $directory = $this->doc
      ->createElement('directory', $this->getPath());
    $exclude = $this->doc
      ->createElement('exclude');
    $exclude_directory = $this->doc
      ->createElement('directory', "{$this->getPath()}/tests");
    $exclude->appendChild($exclude_directory);
    $this->xpath
      ->query('//phpunit/filter/whitelist')
      ->item(0)
      ->appendChild($directory);
    $this->xpath
      ->query('//phpunit/filter/whitelist')
      ->item(0)
      ->appendChild($exclude);
  }
```

PR #182, opened 2021-11-30
Protect against missing includes in example code

danepowell:
In our example module, if the path to the ORCA include is not correct an error will be generated but tests will not necessarily fail.

### DIFF
--- a/example/bin/ci/example.sh
+++ b/example/bin/ci/example.sh
@@ -21,7 +21,7 @@
 cd "$(dirname "$0")" || exit 1
 
 # Reuse ORCA's own includes for its $PATH additions and environment variables.
-source ../../../orca/bin/ci/_includes.sh
+source ../../../orca/bin/ci/_includes.sh || exit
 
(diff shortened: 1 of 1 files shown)


example/bin/ci/example.sh
```diff
@@ -21,7 +21,7 @@
 cd "$(dirname "$0")" || exit 1
 
 # Reuse ORCA's own includes for its $PATH additions and environment variables.
-source ../../../orca/bin/ci/_includes.sh
+source ../../../orca/bin/ci/_includes.sh || exit
```
TravisCarden, line 24: We should go the rest of the way and explicitly emit an error code:

```suggestion
source ../../../orca/bin/ci/_includes.sh || exit 1
```

danepowell: Nit, but I don't think an explicit error code is necessary, since `exit` returns the exit code from the previous command, which in this case is `source`. But it doesn't really matter unless ORCA has a set of explicitly-defined exit codes that I'm not aware of 😄

PR #186, opened 2021-12-15
ORCA jobs not using latest Acquia CMS

danepowell:
A dependency conflict is causing jobs to pull older versions of Acquia CMS. A few things are necessary to fix this:

- Switch from BLT 12 to 13. BLT 12 is EOL and requires older versions of some dependencies that conflict.
- Add Acquia CMS to ORCA's packages.yml. Otherwise the `composer require` command pulls an older version of Acquia CMS. Not sure why, possibly because some dependencies (cohesion) are not "whitelisted" as part of the require command.
- Update ORCA's Pipeline scripts, which were still requiring Lightning and conflicting with Acquia CMS.

See this running in downstream projects:

- https://github.com/acquia/drupal-recommended-project/pull/53
- https://github.com/acquia/blt/pull/4441


config/packages.yml
```diff
@@ -88,17 +92,13 @@ acquia/blt:
   type: composer-plugin
   core_matrix:
     '>=9.0.0':
-      version: 12.x
+      version: 13.x
+      # Yeah I know, it's weird that the dev version is lower than stable. It's because of a stale branch alias.
```
TravisCarden, line 96: Lol. That's funny, but it's written in a different voice from the rest of the codebase. 🙂  And I think we can clarify the terms a little. (They confused me at first.)
```suggestion
      # It doesn't seem right, but "dev_version" must use a lower major version branch than "version" due to a stale branch alias.
```

src/Domain/Fixture/FixtureCreator.php
```diff
@@ -257,6 +257,9 @@ private function fixDefaultDependencies(): void {
       $additions[] = 'phpunit/phpunit:9.4.3';
     }
 
+    // Hack to support Acquia CMS.
```
TravisCarden, line 260: We should have a `@todo` comment in here so we don't lose track of it. If we create a follow-up issue, we can add/replace it with an `@see`.
```diff
-    // Hack to support Acquia CMS.
+    // Hack to support Acquia CMS. Any more detail?
+    // @todo Remove once... ???
```

TravisCarden (changes requested):
A few little things, @danepowell.

src/Domain/Fixture/FixtureCreator.php
danepowell: Added a todo.

TravisCarden (approved):

PR #189, opened 2021-12-17
Don't require Acquia CMS in D8

danepowell:
Latest LTS jobs are failing because ORCA tries to require Acquia CMS in D8. You can see this PR fixes the problem in an actual project: https://github.com/acquia/drupal-environment-detector/actions/runs/1593035947

This wasn't caught because ORCA's internal tests override package.yml to only require drupal/example.

### DIFF
--- a/config/packages.yml
+++ b/config/packages.yml
@@ -49,7 +49,12 @@
 
 acquia/acquia_cms:
   type: drupal-profile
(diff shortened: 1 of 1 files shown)


danepowell:
This shouldn't be an issue now that we aren't running D8 tests.

PR #196, opened 2022-01-14
Trust Composer plugins

danepowell:
[Composer 2.2 requires plugins to be explicitly trusted](https://blog.packagist.com/composer-2-2/).

Without the `allow-plugins` config, users will be prompted to trust the plugins every time they run `composer install` using Composer 2.2+, and non-interactive installs (i.e. in CI) may fail completely.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,6 +66,13 @@
         "phpunit/phpunit": "^9"
     },
     "config": {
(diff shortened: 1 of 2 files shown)


TravisCarden:
Duplicate of https://github.com/acquia/orca/pull/198.

PR #206, opened 2022-03-02
ORCA-333: Provide ability to disable sub modules based on alterations of packages list and disabling drupal/acsf_sso module

secretsayan:
Replaces: https://github.com/acquia/orca/pull/205

### DIFF
--- a/config/packages.yml
+++ b/config/packages.yml
@@ -36,9 +36,9 @@
 #   core version constraint is satisfied. Mappings are processed in order, and
 #   the first match wins.
 #   @see \Acquia\Orca\Tests\Fixture\PackageTest::testConditionalVersions
-# - "enable": (internal) true if the package is a Drupal module that should be
-#   automatically enabled or false if not. Defaults to true for modules. Always
-#   false for anything else.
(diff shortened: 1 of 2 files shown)


config/packages.yml
```diff
@@ -36,9 +36,9 @@
 #   core version constraint is satisfied. Mappings are processed in order, and
 #   the first match wins.
 #   @see \Acquia\Orca\Tests\Fixture\PackageTest::testConditionalVersions
-# - "enable": (internal) TRUE if the package is a Drupal module that should be
-#   automatically enabled or FALSE if not. Defaults to TRUE for modules. Always
-#   FALSE for anything else.
+# - "enable": (internal) true if the package is a Drupal module that should be
+#   automatically enabled or false if not. Defaults to true for modules. Always
+#   false for anything else.
```
TravisCarden, line 41: I realized that this comment isn't entirely accurate anymore.
```suggestion
# - "enable": Whether to enable the extension in Drupal. Defaults to true for
#   modules but can be overridden. Anything else is always false and cannot be
#   overridden.
```

PR #210, opened 2022-03-08
Display JSON error if Drush JSON output is invalid

mglaman:


TravisCarden:
Ah, I like this. Merging.

PR #220, opened 2022-03-28
ORCA-350: Log JUnit output along with coverage

secretsayan:


src/Domain/Tool/Phpunit/PhpUnitTask.php
```diff
@@ -267,6 +267,7 @@ protected function runPhpUnit(): void {
       ];
       if ($this->shouldGenerateCodeCoverage()) {
         $command[] = "--coverage-clover={$this->cloverCoverage}";
+        $command[] = "--log-junit={$this->junitLog}";
       }
```
mglaman, line 271: I wouldn't consider the following comment a block, just bringing it up:

JUnit provides a way to report test results. This may be useful for CI systems which can report these results on jobs not running coverage. However, I think that's an edge case which can be opened later. GHA reads PHPUnit output directly, anyways.

mglaman (approved):

src/Domain/Tool/Phpunit/PhpUnitTask.php
secretsayan: Can you please clarify what do you mean by "report test results"?

mglaman: That's the point of a JUnit XML. It's a standard way to report test results and display them beyond reading `stdout`.

PR #221, opened 2022-03-29
ORCA-349: Code coverage should only be reported against the SUT

secretsayan:


src/Domain/Tool/Phpunit/PhpUnitTask.php
```diff
@@ -132,21 +132,49 @@ private function setTestSuite(): void {
...
+
+    $exclude_directory = $this->doc
+      ->createElement('directory', '../*/contrib/*/tests');
+    $exclude->appendChild($exclude_directory);
+
+    // Appending the excluded directories to "whitelist" element.
+    $whitelist->appendChild($exclude);
+
+    // Adding suffixes to "whitelist" element.
+    $suffixes = ['.php', '.inc', '.module', '.install'];
```
mglaman, line 165: Since this can be used for themes or profiles, we should add:

* `.theme`
* `.profile`
* `.engine`

```suggestion
    // Adding suffixes to "whitelist" element.
    $suffixes = ['.php', '.inc', '.module', '.install', '.theme', '.profile', 'engine'];
```

Here's the list from phpstan-drupal, which I think the additions would cover

```
	fileExtensions:
		- module
		- theme
		- inc
		- install
		- profile
		- engine
```

TravisCarden (changes requested):

PR #223, opened 2022-03-30
ORCA-348: Run PHPUnit with Xdebug in coverage mode when ORCA_COVERAGE_ENABLE

secretsayan:


src/Helper/Process/ProcessRunner.php
```diff
@@ -226,6 +251,13 @@ protected function createVendorBinProcess(array $command): Process {
...
     }
 
+    if ($this->isEnvVarsSet === TRUE) {
+      $command = array_merge([
+        $this->envVars,
+        "php",
+      ], $command);
+    }
+
     return new Process($command);
```
TravisCarden, line 261: The Symfony Process component actually has special handling for environment variables. See [Setting Environment Variables for Processes](https://symfony.com/doc/current/components/process.html#setting-environment-variables-for-processes). It might look something like this, for example:

```suggestion
    return new Process($command, null, $this->envVars);
```

Of course, that assumes that `::envVars` is an array with an initial value of `[]`. Then you don't have to have `::isEnvVarsSet` at all. And you can replace `::setEnvVars()` with something a little resilient/type-safe like this, for example:

```
public function setEnv(string $name, string $value): void {
  $this->envVars[] = [$name => $value];
}
```

mglaman (approved):
This looks like it should work! Is there a CI job which runs ORCA on the test repo with coverage, so we can verify?

src/Helper/Process/ProcessRunner.php
```diff
@@ -211,6 +218,18 @@ public function createOrcaVendorBinProcess(array $command): Process {
...
+   * Set environment variables.
+   *
+   * @param string $name
+   *   The name of the environment variables required to be set.
+   * @param string $value
+   *   The value of the environment variables required to be set.
+   */
+  public function setEnvVars(string $name, string $value): void {
+    $this->envVars[$name] = $value;
+  }
```
TravisCarden, line 231:
```suggestion
  /**
   * Adds an environment variable.
   *
   * @param string $name
   *   The name of the environment variable.
   * @param string $value
   *   The value of the environment variable.
   */
  public function addEnvVar(string $name, string $value): void {
    $this->envVars[$name] = $value;
  }
```

src/Helper/Process/ProcessRunner.php
```diff
@@ -36,6 +36,13 @@ class ProcessRunner {
    */
   private $output;
 
+  /**
+   * Array of Environment Variables.
```
TravisCarden, line 40:
```suggestion
   * An array of environment variables.
```

src/Domain/Tool/Phpunit/PhpUnitTask.php
```diff
@@ -267,6 +267,7 @@ protected function runPhpUnit(): void {
       ];
       if ($this->shouldGenerateCodeCoverage()) {
         $command[] = "--coverage-clover={$this->cloverCoverage}";
+        $this->processRunner->setEnvVars("XDEBUG_MODE", "coverage");
```
TravisCarden, line 270:
```suggestion
        $this->processRunner->addEnvVar("XDEBUG_MODE", "coverage");
```

TravisCarden (commented):
Making a couple of little changes...

PR #227, opened 2022-04-13
Add all-successful job

danepowell:
GitHub branch protection requires explicitly defining each job that must pass in order to allow merging a PR. The problem is that ORCA's jobs are defined dynamically via a matrix, meaning that to use protected branches you need to copy every single job name into the branch protection settings and constantly keep them updated as ORCA is updated.

A much better solution is to define a single "all-successful" job that only passes if all other ORCA jobs pass, and then you can configure branch protection to just require "all-successful".

### DIFF
--- a/example/.github/workflows/orca.yml
+++ b/example/.github/workflows/orca.yml
@@ -216,3 +216,11 @@ jobs:
       - name: After failure
         if: ${{ failure() }}
         run: ../orca/bin/ci/after_failure.sh
+
+  all-successful:
+    # https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert
+    runs-on: ubuntu-latest
+    needs: [build]
+    steps:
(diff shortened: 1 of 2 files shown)


example/.github/workflows/orca.yml
```diff
@@ -216,3 +216,11 @@ jobs:
       - name: After failure
         if: ${{ failure() }}
         run: ../orca/bin/ci/after_failure.sh
+
+  all-successful:
+    # https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert
+    runs-on: ubuntu-latest
+    needs: [build]
+    steps:
+    - name: note that all tests succeeded
```
TravisCarden, line 225:
```suggestion
    - name: All checks successful
```

TravisCarden (changes requested):

TravisCarden (approved):

PR #231, opened 2022-04-25
Remove symfony/http-kernel and upgrade composer to latest version

secretsayan:


var/log/phpcs.json
```diff
@@ -0,0 +1 @@
+
```
TravisCarden, line 1: Looks like contents of the `var` directory have been committed somehow. Please remove the newly-committed files and figure out if something was broken to allow them to get committed in the first place.

bin/orca.php
```diff
@@ -19,35 +15,37 @@
...
+$container->setParameter('app.project_dir', dirname(__DIR__));
+$container->setParameter('app.fixture_dir', dirname(__DIR__) . '/../orca-build');
+
+$container->compile(TRUE);
+
+/**
+ * Console application class.
+ *
+ * @var \Symfony\Component\Console\Application $application
+ */
```
TravisCarden, line 38: I feel like the comment is unnecessary here..

```suggestion
/** @var \Symfony\Component\Console\Application $application */
```

composer.json
```diff
@@ -25,45 +25,45 @@
         "ext-dom": "*",
         "ext-json": "*",
         "ext-sqlite3": "*",
-        "acquia/coding-standards": "^0.7.0",
-        "composer/composer": "~2.2.5",
+        "acquia/coding-standards": "^0.8.0",
```
TravisCarden, line 28: Oops. This is probably technically a backward-compatibility break that will need communication beforehand. Would you create an issue in our backlog to do the communication and then the update? In the meantime, lets drop this back down.

```suggestion
        "acquia/coding-standards": "^0.7.0",
```

composer.json
```diff
@@ -25,45 +25,45 @@
...
+        "composer/composer": "~2.3.5",
         "cweagans/composer-patches": "^1.7",
-        "dealerdirect/phpcodesniffer-composer-installer": "^0.7.1",
-        "ergebnis/composer-normalize": "^2.0.0",
-        "hassankhan/config": "^2.1",
+        "dealerdirect/phpcodesniffer-composer-installer": "^0.7.2",
+        "ergebnis/composer-normalize": "~2.15.0",
+        "hassankhan/config": "^3.0",
```
TravisCarden, line 33: I'm sure this is the source of the merge conflict ([ #235 ](https://github.com/acquia/orca/pull/235)). 🙂

composer.json
```diff
@@ -25,45 +25,45 @@
...
-        "phpmd/phpmd": "^2.6",
-        "symfony/config": "^4.1",
-        "symfony/console": "^4.1",
-        "symfony/dependency-injection": "^4.1",
+        "php-coveralls/php-coveralls": "^2.5",
+        "php-parallel-lint/php-console-highlighter": "^1.0",
+        "php-parallel-lint/php-parallel-lint": "^1.3",
+        "phploc/phploc": "^7.0",
+        "phpmd/phpmd": "^2.12",
+        "stecman/symfony-console-completion": "^0.11.0",
```
TravisCarden, line 42: I distinctly remember us removing this feature from `orca.php` and documentation. It looks like it's getting _added_ here, not just updated. Let's remove it.

And while you're at it, please compare the additions to this file with the removals to make sure we didn't accidentally (re-)introduce any extraneous dependencies.

composer.json
```diff
@@ -25,45 +25,45 @@
...
         "symfony/symfony": "*"
     },
     "require-dev": {
-        "brainmaestro/composer-git-hooks": "^2.5",
+        "brainmaestro/composer-git-hooks": "^2.8",
         "phan/phan": "^5.3",
         "phpspec/prophecy-phpunit": "^2.0",
-        "phpstan/phpstan": "^1.4",
-        "phpunit/phpunit": "^9"
+        "phpstan/phpstan": "^1.5",
```
TravisCarden, line 65: Please make sure there are no breaking changes in this version of PHPStan. In theory, a minor version update shouldn't have any, but if it did, it would cause a lot of headaches.

composer.json
```diff
@@ -85,7 +85,7 @@
             "pre-commit": [
                 "set -e",
                 "vendor/bin/parallel-lint --exclude vendor --no-progress .",
-                "vendor/bin/phpcs --cache=var/cache/phpcs.json"
+                "vendor/bin/phpcs --no-cache"
```
TravisCarden, line 88: I don't think there's any need to change this, actually.

```suggestion
                "vendor/bin/phpcs --cache=var/cache/phpcs.json"
```

composer.json
```diff
@@ -123,7 +123,7 @@
         "open-coverage": "open var/coverage-report/index.html",
         "phan": "vendor/bin/phan -k phan.php --allow-polyfill-parser",
         "phpcbf": "phpcbf",
-        "phpcs": "phpcs -s --cache=var/cache/phpcs.json",
+        "phpcs": "phpcs -s",
```
TravisCarden, line 126: Ditto above:

```suggestion
        "phpcs": "phpcs -s --cache=var/cache/phpcs.json",
```

composer.json
```diff
@@ -25,45 +25,45 @@
...
+        "symfony/filesystem": "^5.4",
+        "symfony/finder": "^5.4",
+        "symfony/http-client": "^5.4",
+        "symfony/options-resolver": "^5.4",
+        "symfony/phpunit-bridge": "^6.0",
+        "symfony/process": "^5.4",
+        "symfony/yaml": "^5.4",
         "webdriver-binary/binary-chromedriver": "^6.1.0",
-        "weitzman/drupal-test-traits": "^1.3",
+        "weitzman/drupal-test-traits": "^1.5",
```
TravisCarden, line 55: I also seem to remember changing the way we add `weitzman/drupal-test-traits` to fixtures. Please determine whether we need to `require` it here at all.

config/services.yml
```diff
@@ -4,13 +4,13 @@ parameters:
...
-  env(ORCA_COVERAGE_CLOVER): "%kernel.project_dir%/var/coverage/clover.xml"
+  env(ORCA_COVERAGE_CLOVER): "%app.project_dir%/var/coverage/clover.xml"
   env(ORCA_FIXTURE_DIR): "%app.fixture_dir%"
-  env(ORCA_JUNIT_LOG): "%kernel.project_dir%/var/junit/junitLog.xml"
+  env(ORCA_JUNIT_LOG): "%app.project_dir%/var/junit/junitLog.xml"
   env(ORCA_PACKAGES_CONFIG): config/packages.yml
   env(ORCA_PACKAGES_CONFIG_ALTER): ~
   env(ORCA_PHPCS_STANDARD): "AcquiaDrupalTransitional"
-  env(ORCA_TELEMETRY_ENABLE): false
+  env(ORCA_TELEMETRY_ENABLE): "false"
```
TravisCarden, line 13: Did you check to make sure our environment variable reading functionality handles this identically? (In the absence of the variable on the shell and in its presence--with both `true` and `false` values?

TravisCarden (changes requested):

config/services.yml
secretsayan: I have checked with the following combinations in my local-
- export ORCA_TELEMETRY_ENABLE=true
- export ORCA_TELEMETRY_ENABLE=false
- unset ORCA_TELEMETRY_ENABLE

All these behave in the same way before and after the change

PR #232, opened 2022-04-26
Fix version resolution logic of NEXT_MAJOR_LATEST_MINOR_BETA

secretsayan:


src/Domain/Composer/Version/DrupalCoreVersionResolver.php
```diff
@@ -400,4 +403,20 @@ private function convertToDev(string $version): string {
     return implode('.', $parts);
   }
 
+  /**
+   * Validates if the version is "NEXT MAJOR LATEST MINOR BETA OR LATER".
```
TravisCarden, line 407: Is this quite right? I feel like it should be something more like "Asserts that a next major, latest minor beta-or-later Drupal core version exists."

src/Domain/Composer/Version/DrupalCoreVersionResolver.php
```diff
@@ -400,4 +403,20 @@ private function convertToDev(string $version): string {
...
+   *
+   * @param string $version
+   *   The version to validate.
+   *
+   * @throws \Acquia\Orca\Exception\OrcaVersionNotFoundException
+   */
+  private function validateNextMajorLatestMinorBetaOrLater($version): void {
+    $parts = explode('.', $version);
+    $stability = end($parts);
+    if ((strpos($stability, "alpha") !== FALSE) || (strpos($stability, "dev") !== FALSE)) {
```
TravisCarden, line 417: Let's have some more [good, expressive variable names](https://wiki.c2.com/?GoodVariableNames=) to make the logic clearer here. For example:

```php
$is_alpha = strpos($stability, "alpha") !== FALSE;
```

src/Domain/Composer/Version/DrupalCoreVersionResolver.php
```diff
@@ -400,4 +403,20 @@ private function convertToDev(string $version): string {
...
+      throw new OrcaVersionNotFoundException();
```
TravisCarden, line 418: Let's include an exception message. See `::findLatestLts()` for comparison:

```php
    $message = "No Drupal core version satisfies the given constraints: oldest supported ($oldestSupported) less than current major ($current_major)";
    throw new OrcaVersionNotFoundException($message);
```

TravisCarden (changes requested):
Looks good! Just a few little suggestions.

TravisCarden (approved):

PR #244, opened 2022-05-30
Install dev dependencies of the SUT during fixture creation

secretsayan:


src/Domain/Fixture/FixtureCreator.php
```diff
@@ -437,6 +439,66 @@ private function verifySut(): void {
...
+    }
+    $subextensions = $this->subextensionManager->getByParent($package);
+
+    do {
+      $dev_dependencies[] = $this->computeDevDependenciesByPackage($package);
+      if (count($subextensions) <= 0) {
+        break;
+      }
+      $package = array_shift($subextensions);
+    } while (1);
```
TravisCarden, line 468: Let's use a `foreach` loop here. It's much simpler and easier to read. Just eyeballing it, something like this, I think:

```suggestion
foreach ($packages as $package) {
  $dev_dependencies[] = $this->computeDevDependenciesByPackage($package);
}
```

src/Domain/Fixture/FixtureCreator.php
```diff
@@ -437,6 +439,66 @@ private function verifySut(): void {
...
+   *   List of dev-dependencies found.
+   *
+   * @throws \Acquia\Orca\Exception\OrcaException
+   * @throws \Acquia\Orca\Exception\OrcaFileNotFoundException
+   * @throws \Acquia\Orca\Exception\OrcaParseError
+   */
+  private function computeDevDependenciesByPackage(Package $package): ?string {
+    $dev_dependencies =
+      $this->subextensionManager->findDevDependenciesByPackage($package);
+    if ($dev_dependencies === NULL || count($dev_dependencies) === 0) {
```
TravisCarden, line 490: I think this would be just slightly clearer:

```suggestion
    if ($dev_dependencies === NULL || $dev_dependencies === []) {
```

src/Domain/Fixture/SubextensionManager.php
```diff
@@ -180,6 +181,30 @@ public function getByParent(Package $package): array {
...
+   *   The package to search for dev-dependencies.
+   *
+   * @return array|null
+   *   An indexed array of all the dev-dependencies.
+   *
+   * @throws \Acquia\Orca\Exception\OrcaException
+   * @throws \Acquia\Orca\Exception\OrcaFileNotFoundException
+   * @throws \Acquia\Orca\Exception\OrcaParseError
+   */
+  public function findDevDependenciesByPackage(Package $package): ?array {
```
TravisCarden, line 197: If you _always_ return an array--even an empty one--then code that uses this function can safely just iterate over it without have to check first that it's an array. For example, the code that calls this above:

```diff
- if ($dev_dependencies === NULL || $dev_dependencies === []) {
+ if ($dev_dependencies === []) {
```

secretsayan: Done

TravisCarden (approved):
Looks good!

PR #251, opened 2022-06-13
ORCA-391: all-successful job doesn't catch build failures

danepowell:
This fixes the issue by requiring that the all-successful job run always and using the re-actors action to evaluate whether required jobs passed.

You can see proof that this works here: https://github.com/acquia/cli/pull/1040

And here's an implementation in an actual ORCA project: https://github.com/acquia/blt/pull/4539


TravisCarden:
Cool, @danepowell! @secretsayan, let's talk about whether or not we need to send a notification to ORCA users or even pull requests.

PR #252, opened 2022-06-13
Auto-merge dependency updates

danepowell:
I see you've enabled dependabot on this repo. I find manual review of dependabot PRs to be onerous and unnecessary. If you have good test coverage and tests pass, you might as well just automatically merge the PR and save yourself a few clicks.

I use this on all of my other repos (Acquia CLI, BLT, etc), it works great! The only difference is that we use Violinist and you use Dependabot, but I think it should still work.

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -1,0 +1,28 @@
+name: Merge dependency updates
+on:
+  workflow_run:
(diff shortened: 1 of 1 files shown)


TravisCarden:
Cool. Thanks, @danepowell!

PR #253, opened 2022-06-16
Update static analysis tools for upstream Composer allow-plugins changes

secretsayan:


src/Domain/Composer/ComposerFacade.php
```diff
@@ -261,7 +261,7 @@ public function validate(string $path): void {
       '--ansi',
       'validate',
     ];
-    $this->runComposer($command, [$path], $this->orca->getPath());
+    $this->runComposer($command, [$path], dirname($this->fixture->getPath()));
```
TravisCarden, line 264: Why do we need to set the CWD to the directory _above_ the fixture root, @secretsayan? Is it just so Composer isn't interpreting the fixture's Composer config (i.e., it's `composer.json`)? If it doesn't actually matter, we don't need to pass a value here at all, because it defaults to the fixture root. If it does matter, would it work like the below? Just from memory, I think `::getPath()` might support that, and it would be preferable to using a system function directory--for unit testing, for example.

```php
$this->fixture->getPath('..')
```

secretsayan: Modified the code as per suggestions.

Yes, this path does matter, otherwise it would interfere with the composer.json config present in the fixture itself.

src/Domain/Fixture/FixtureCreator.php
```diff
@@ -316,6 +316,11 @@ private function addAllowedComposerPlugins(): void {
       }
     }
 
+    // Adding this to allowed-plugins, to counter the error in our integrated
+    // builds. This should be removed when drupal-recommended-project is
+    // re-enabled using PR https://github.com/acquia/orca/pull/257.
```
TravisCarden, line 321:
```suggestion
    // @todo Adding this to allowed-plugins, to counter the error in our
    //   integrated builds. This should be removed when
    //   drupal-recommended-project is re-enabled using PR
    //   https://github.com/acquia/orca/pull/257.
```

PR #259, opened 2022-07-05
Add exception handling to composer normalize

secretsayan:


src/Domain/Composer/ComposerFacade.php
```diff
@@ -282,7 +285,12 @@ public function normalize(string $path, array $args = []): void {
     $command = array_merge($command, $args);
     // The cwd must be the ORCA project directory in order for Composer to
     // find the "normalize" command.
-    $this->runComposer($command, [$path], $this->orca->getPath());
+    try {
+      $this->runComposer($command, [$path], $this->orca->getPath());
+    }
+    catch (\Exception $e) {
+      throw new OrcaException($e->getMessage());
```
TravisCarden, line 292: The point here is to suppress Composer Normalizer failures, right, @secretsayan? Wouldn't throwing a new exception still let it fail?

secretsayan: @TravisCarden: I was thinking that this [line](https://github.com/acquia/orca/blob/develop/src/Domain/Tool/ComposerValidate/ComposerValidateTask.php#L53) would suppress the `OrcaException`. Anyways, I have made the necessary changes.

secretsayan:
This PR may be closed as the changes are moved to https://github.com/acquia/orca/pull/253

PR #261, opened 2022-07-07
Update to Acquia CMS 2.0

secretsayan:


secretsayan:
@vishalkhode1 : It seems ACMS 2.0 has some conflict with lightning_page. can you please check and suggest

cc: @TravisCarden

TravisCarden (approved):

PR #262, opened 2022-07-08
ORCA-400: Fix Composer Normalize bug

secretsayan:


TravisCarden (changes requested):
@secretsayan, is there any way to accomplish this without resorting to global config? It's best not to have side effects outside the fixture, and we've successfully avoided that so far. Could we perhaps invoke the normalizer via its code API so the plugin never has to be loaded anywhere at all?

secretsayan:
Closing this PR as this will be automatically resolved in the next release of Composer, which is `composer 2.3.11`

PR #273, opened 2022-08-12
Update before_install.sh

secretsayan:


secretsayan:
This is not required anymore as pecl is working fine

PR #274, opened 2022-08-17
ORCXA-422: Fix YamlLint 

secretsayan:


wimleers:
There's no information provided here.

What problem did this fix?

This introduced a significant regression: `.yml` files that are not targeted at Drupal are validated too. For example, those associated with a CI system in a `.hidden` directory.

PR #277, opened 2022-08-25
Install and use Chromedriver on the host to fix timeouts

danepowell:
Both Chrome and Chromedriver should already be available OOTB on the devops and GitHub Actions pipelines. I don't know why we're still supporting Travis CI.


TravisCarden (approved):
This looks good to me!

PR #278, opened 2022-08-25
Enable dependabot

danepowell:
Some of ORCA's dependencies are outdated and actually have breaking changes that need to be addressed. It's good to keep these updated automatically.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "composer"
+    directory: "/"
+    schedule:
+      interval: "daily"


composer.json
```diff
@@ -72,6 +72,9 @@
             "webdriver-binary/binary-chromedriver": true
         },
         "discard-changes": true,
+        "platform": {
+            "php": "7.4"
```
danepowell, line 76: This prevents Dependabot (or anyone else for that matter) from accidentally installing dependencies that won't run in PHP 7.4

TravisCarden: A change like this should have its own PR (and discussion 🙂). Dependabot can just specify its PHP version in its YAML file, right?

danepowell: No, as far as I know this is the only way to constrain the PHP version.

Why would you not want this? If you install a PHP 8 dependency ORCA's tests will fail (as happened to me on #277 initially, what a pain). Limiting packages to those that support PHP 7.4 can only be a good thing.

danepowell:
@secretsayan can someone from your team review this? I think you'd be happier with automated dependency updates, but please either close this or merge it 😄 

If you decide to close it, you should delete the automerge workflow and anything else related to automated updates.

TravisCarden:
@secretsayan, we can discuss this any time you like, and then I intend to leave the final decision up to you.

composer.json
secretsayan: @danepowell : ORCA needs to be installable in php 8.1 as well, so can we remove the platform config mentioned above. 

Also as we need to install ORCA in multiple php versions (7.4, 8.0 and 8.1) we may not pin our dependencies to a specific php version. We need to depend on CI finding it and what happened to you is exactly what we want to happen.

secretsayan:
Thanks @danepowell.

secretsayan (approved):

TravisCarden (approved):

PR #279, opened 2022-08-26
Make jobs work on Drupal 10

secretsayan:


danepowell:
I would suggest not removing config platform universally, but only for jobs that you _know_ conflict with it, i.e. Drupal 10.

For instance, if the config platform for some reason caused Drupal 9 installs to fail, that's an error that we would want ORCA to catch.

TravisCarden:
@danepowell, wouldn't removing `config.platform.php` _only_ for Drupal 10 conflict with https://github.com/acquia/drupal-recommended-project/pull/220, where you're trying to increase it to 8.0? It seems like that would cause all of our PHP 7.4 jobs to fail. Am I missing something?

danepowell:
I guess "jobs we know to fail" would then include Drupal 10 jobs and PHP 7.4 jobs.

If it's easier to just remove the config from every job, I won't complain.

src/Domain/Fixture/FixtureCreator.php
```diff
@@ -211,6 +212,19 @@ private function createComposerProject(): void {
...
+  /**
+   * Remove "config.platform" parameter from fixture root composer.json.
+   */
+  private function removeComposerConfigPlatform(): void {
+    $this->output->writeln("Removing Platform Dependencies");
+    try {
+      $this->composer->removeConfig(['platform']);
+    }
+    catch (\Exception $e) {
+      $this->output->writeln("Operation failed.");
```
TravisCarden, line 224: Let's be a little more specific with our error message, e.g.,

```suggestion
      $this->output->writeln("Failed to remove Composer platform requirements.");
```

We should update the message on line 219 accordingly, e.g., "Removing Composer platform requirements". And it wouldn't hurt to rename the method to match.

src/Domain/Fixture/FixtureCreator.php
```diff
@@ -252,8 +266,11 @@ private function fixDefaultDependencies(): void {
     }
 
     // Acquia CMS uses drupal-test-traits as a dev dependency.
+    // Add only when drupal core version is NOT 10.
```
TravisCarden, line 269:
```suggestion
    // Add only when Drupal core version is NOT 10.
```

src/Domain/Fixture/SiteInstaller.php
```diff
@@ -187,8 +187,7 @@ private function addDependencies(): void {
       'help',
       'toolbar',
     ]);
-    $this->addItems($info_file, 'themes', ['seven']);
-
+    $this->addItems($info_file, 'themes', ['claro']);
```
TravisCarden, line 190:
```suggestion
    $this->addItems($info_file, 'themes', ['claro']);

```
🙂

src/Domain/Tool/Phpunit/PhpUnitTask.php
```diff
@@ -143,13 +143,25 @@ private function setTestSuite(): void {
...
+      $whitelist = $this->doc->createElement('whitelist');
+    }
+    else {
+      // Checking Drupal 10 case.
+      $whitelist = $this->xpath->query('//phpunit/coverage')->item(0);
+      assert($whitelist instanceof \DOMElement);
+      $whitelist->parentNode->removeChild($whitelist);
+      $appendTo = "//phpunit";
+
+      // Creating new "whitelist" element.
```
TravisCarden, line 162:
```suggestion
      // Creating new "coverage" element.
```

src/Options/FixtureOptions.php
```diff
@@ -343,6 +341,27 @@ public function getCoreResolved(): string {
     return $version;
   }
 
+  /**
+   * Check if core version resolves to the constraint provided.
```
TravisCarden, line 345:
```suggestion
   * Checks if core version resolves to the constraints provided.
```

I don't make the rules. 🤷 😉

> For most functions... summary lines must start with a third person singular present tense verb, and they must explain what the function does. Example: "Calculates the maximum weight for a list."

\- https://www.drupal.org/docs/develop/standards/api-documentation-and-comment-standards#functions

src/Options/FixtureOptions.php
```diff
@@ -343,6 +341,27 @@ public function getCoreResolved(): string {
...
+   *
+   * @param string $constraints
+   *   The constraint to check for.
```
TravisCarden, line 348:
```suggestion
   * @param string $constraints
   *   The constraints to check for.
```

tests/Domain/Composer/ComposerFacadeTest.php
```diff
@@ -257,6 +257,34 @@ public function testIsValidConstraint($expected, $version): void {
     self::assertEquals($expected, $actual, 'Correctly determined validity of version constraint.');
   }
 
+  public function configList(): array {
```
TravisCarden, line 260: This is the convention in the rest of the tests:
```suggestion
  public function providerRemoveConfig(): array {
```
Also, the provider always goes _after_ the test method.

tests/Domain/Composer/ComposerFacadeTest.php
```diff
@@ -257,6 +257,34 @@ public function testIsValidConstraint($expected, $version): void {
...
+    return [
+      [['platform']],
+    ];
+  }
+
+  /**
+   * @dataProvider configList
```
TravisCarden, line 267: ...and, correspondingly:

```suggestion
   * @dataProvider providerRemoveConfig
```

TravisCarden (changes requested):
Nice. A handful of minor points.

src/Options/FixtureOptions.php
secretsayan: Awesome :)

TravisCarden (approved):

PR #280, opened 2022-08-30
Ensure that the platform Composer bin is used on CI

secretsayan:


secretsayan:
Proof that this works with ACQUIA CI https://github.com/acquia/devops-orca/pull/22

PR #284, opened 2022-09-19
ORCA-419: Add Gitlab OAuth token

secretsayan:


TravisCarden:
Hmm. If we took this approach, every product would have to configure their own private token, right? I can't see that working well for us; I wonder if there's another way. We're doing this to get around rate limits, as I recall? If we require a specific version of `weitzman/drupal-test-traits`, e.g., `1.6.0`, would that reduce the number of requests enough to circumvent the problem?

secretsayan:
Closing this pull request as we cannot expose personal access tokens in our code.

PR #288, opened 2022-09-28
Change precedence of PATH values for finding correct version of PHP

secretsayan:
Change the order of the exported PATH variable to aid in choosing the correct version of php in CI systems.


TravisCarden (changes requested):
This looks great, if you can figure out why the tests are failing, @secretsayan.

secretsayan:
@TravisCarden : I re-ran the tests and they are passing now.

TravisCarden (approved):

PR #293, opened 2022-10-14
ORCA-457: Pin drush to a stable version 

secretsayan:


secretsayan:
This change is not required anymore as the issue has been fixed upstream.

PR #294, opened 2022-10-14
ORCA-452: Fix Drupal Recommended Project tests

secretsayan:


src/Domain/Composer/ComposerFacade.php
```diff
@@ -131,6 +131,7 @@ public function createProjectFromPackage(Package $package): void {
     ]);
     $this->runComposer([
       'create-project',
+      '--stability=dev',
```
danepowell, line 134: This needs to be conditional. Only use `--stability=dev` if the SUT is a project template.

Otherwise, a project running a stable Drupal 9 job could accidentally install a D10 release of the project template, e.g. 10.0.0-alpha1, the same bug as ORCA-446

danepowell (changes requested):
This does fix DRP: https://github.com/acquia/drupal-recommended-project/pull/245

But I suspect it will cause a recurrence of ORCA-446 if the DRP cuts another D10 alpha release.

src/Domain/Composer/ComposerFacade.php
secretsayan: @danepowell : Actually this function only runs when the SUT _is a_ project template. If the SUT _is not a_ project template ORCA runs a different function. Please check below

_https://github.com/acquia/orca/blob/895edf8c914f5ed8bf6d762cf7e58ba54c3b329d/src/Domain/Fixture/CodebaseCreator.php#L104-L119_

danepowell (approved):
Oh okay then, looks good

TravisCarden (approved):
Looks good to me, @secretsayan! You may merge it.

PR #326, opened 2023-01-16
Point to appropriate version of DRP

secretsayan:
This change will prevent ORCA outages when DRP releases new latest tag for D10 and has to be merged and released before DRP's D10 release.

cc: @vishalkhode1

### DIFF
--- a/src/Options/FixtureOptions.php
+++ b/src/Options/FixtureOptions.php
@@ -300,14 +300,13 @@ class FixtureOptions {
       return $this->options['project-template'];
     }
 
+    // @todo Point to latest version of DRP once 2.x for D10 stable fixtures
+    //   and "dev-master" for D10 dev fixtures.
     if ($this->coreVersionParsedMatches('^10')) {
       $this->options['project-template'] = 'acquia/drupal-recommended-project:dev-drupal10';
     }
-    elseif ($this->coreVersionParsedMatches('^9') && $this->isDev()) {
-      $this->options['project-template'] = 'acquia/drupal-recommended-project:dev-master';
-    }
     else {
-      $this->options['project-template'] = 'acquia/drupal-recommended-project';
+      $this->options['project-template'] = 'acquia/drupal-recommended-project:^1';
     }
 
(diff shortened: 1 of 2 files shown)


src/Options/FixtureOptions.php
```diff
@@ -300,14 +300,13 @@ public function getProjectTemplate(): string {
       return $this->options['project-template'];
     }
 
+    // @todo Point to latest version of DRP once 2.x for D10 stable fixtures
+    //   and "dev-master" for D10 dev fixtures.
     if ($this->coreVersionParsedMatches('^10')) {
       $this->options['project-template'] = 'acquia/drupal-recommended-project:dev-drupal10';
     }
-    elseif ($this->coreVersionParsedMatches('^9') && $this->isDev()) {
-      $this->options['project-template'] = 'acquia/drupal-recommended-project:dev-master';
```
vishalkhode1, line 307: @secretsayan If we want to support D9 dev, we can use [dev-1.x](https://github.com/acquia/drupal-recommended-project/tree/1.x) branch.

secretsayan: Currently we do not have any job that would point to D9 dev.


Screenshot 2023-01-19 at 7 18 52 PM

vishalkhode1: Ok thanks, make sense. 👍

vishalkhode1 (approved):

src/Options/FixtureOptions.php
```diff
@@ -300,14 +300,13 @@ public function getProjectTemplate(): string {
       return $this->options['project-template'];
     }
 
+    // @todo Point to latest version of DRP once 2.x for D10 stable fixtures
```
TravisCarden, line 303: As a rule, we shouldn't use abbreviations unless they're widely known. We have no idea who may be reading this code in the future or whether "DRP" will mean anything to them. "D10", for example, is common enough that we can use it.

TravisCarden (approved):
One FYI. No need to change anything for this merge.

PR #327, opened 2023-01-17
Adding Drupal core nightwatch tests

secretsayan:
We want to run Drupal Core's nightwatch tests so that we can increase the coverage of testing nightwatch in ORCA

### DIFF
--- a/bin/ci/script.sh
+++ b/bin/ci/script.sh
@@ -30,6 +30,15 @@ if [[ "$ORCA_ENABLE_NIGHTWATCH" == "TRUE" && "$ORCA_SUT_HAS_NIGHTWATCH_TESTS" &&
       --passWithNoTests \\
       --tag=$ORCA_SUT_MACHINE_NAME"
 
+    if [[ "$ORCA_SUT_NAME" == "drupal/example" ]]; then
+      (
+        eval "yarn test:nightwatch \\
+          --headless \\
+          --passWithNoTests \\
+          --tag=core"
+      )
+    fi
+
     kill -0 $SERVER_PID
     kill -0 $CHROMEDRIVER_PID
   )
--- a/bin/ci/script.sh
+++ b/bin/ci/script.sh
@@ -30,15 +30,6 @@ if [[ "$ORCA_ENABLE_NIGHTWATCH" == "TRUE" && "$ORCA_SUT_HAS_NIGHTWATCH_TESTS" &&
       --passWithNoTests \\
       --tag=$ORCA_SUT_MACHINE_NAME"
 
-    if [[ "$ORCA_SUT_NAME" == "drupal/example" ]]; then
-      (
-        eval "yarn test:nightwatch \\
-          --headless \\
-          --passWithNoTests \\
-          --tag=core"
-      )
-    fi
-
     kill -0 $SERVER_PID
     kill -0 $CHROMEDRIVER_PID
   )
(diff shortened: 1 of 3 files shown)


bin/ci/script.sh
```diff
@@ -30,6 +30,15 @@ if [[ "$ORCA_ENABLE_NIGHTWATCH" == "TRUE" && "$ORCA_SUT_HAS_NIGHTWATCH_TESTS" &&
       --passWithNoTests \\
       --tag=$ORCA_SUT_MACHINE_NAME"
 
+    if [[ "$ORCA_SUT_NAME" == "drupal/example" ]]; then
```
TravisCarden, line 33: This is internal/self-test code. We don't want it in the customer-facing `script.sh`. It should be in `bin/ci/self-test/script.sh`.

TravisCarden (changes requested):
One change requested.

bin/ci/self-test/script.sh
```diff
@@ -44,3 +44,28 @@ fi
 if [[ "$ORCA_LIVE_TEST" ]]; then
   orca qa:automated-tests
 fi
+
+if [[ "$ORCA_ENABLE_NIGHTWATCH" == "TRUE" && "$ORCA_SUT_HAS_NIGHTWATCH_TESTS" && -d "$ORCA_YARN_DIR" &&  "$ORCA_SUT_NAME" == "drupal/example" ]]; then
```
TravisCarden, line 48: Niggle. 🙂
```suggestion
if [[ "$ORCA_ENABLE_NIGHTWATCH" == "TRUE" && "$ORCA_SUT_HAS_NIGHTWATCH_TESTS" && -d "$ORCA_YARN_DIR" && "$ORCA_SUT_NAME" == "drupal/example" ]]; then
```

bin/ci/script.sh
```diff
@@ -15,7 +15,7 @@ if [[ "$ORCA_JOB" ]]; then
   eval "orca ci:run $ORCA_JOB script $ORCA_SUT_NAME"
 fi
 
-if [[ "$ORCA_ENABLE_NIGHTWATCH" == "TRUE" && "$ORCA_SUT_HAS_NIGHTWATCH_TESTS" && -d "$ORCA_YARN_DIR" ]]; then
+if [[ "$ORCA_ENABLE_NIGHTWATCH" == "TRUE" && "$ORCA_SUT_HAS_NIGHTWATCH_TESTS" && -d "$ORCA_YARN_DIR" && "$ORCA_SUT_NAME" != "drupal/example" ]]; then
```
TravisCarden, line 18: We still have handling for an internal-only (self-test) variable here--`$ORCA_SUT_NAME`--which is what we're trying to avoid. You should remove it from this condition here and then set `$ORCA_ENABLE_NIGHTWATCH` to `FALSE` before running `bin/ci/script.sh` in your self-test CI configuration, e.g., GitHub Actions. Then set it back to `TRUE` before running `bin/ci/self-test/script.sh`. That way it will be disabled in the common script that customers use, without having to be handled in customer-facing code. But it will still work for us.

TravisCarden (changes requested):

bin/ci/script.sh
secretsayan: Done

bin/ci/self-test/script.sh
secretsayan: This is not required as we have removed the check `"$ORCA_SUT_NAME" == "drupal/example" `

TravisCarden (approved):

PR #313, opened 2022-12-21
Remove Travis CI Integration

secretsayan:

PR #328, opened 2023-01-19
ORCA-507: Remove Coveralls

secretsayan:
**Motivation**
- We are removing Coveralls from ORCA as mandated by Acquia
- Remove references from documentation.


docs/getting-started.md
```diff
@@ -12,14 +12,13 @@
...
-* **[`after_success.sh`](../bin/common/after_success.sh)** is reserved for future use.
-* **[`after_failure.sh`](../bin/common/after_failure.sh)** displays debugging information in case of failure.
-* **[`after_script.sh`](../bin/common/after_script.sh)** conditionally logs the job.
+* **[`before_install.sh`](../bin/ci/before_install.sh)** configures the Travis CI environment, installs ORCA, and prepares the SUT.
+* **[`install.sh`](../bin/ci/install.sh)** creates the test fixture and places the SUT.
+* **[`before_script.sh`](../bin/ci/before_script.sh)** displays details about the fixture for debugging purposes.
+* **[`script.sh`](../bin/ci/script.sh)** runs static code analysis and automated tests.
+* **[`before_cache.sh`](../bin/ci/before_cache.sh)** is reserved for future use.
+* **[`after_success.sh`](../bin/ci/after_success.sh)** is reserved for future use.
+* **[`after_failure.sh`](../bin/ci/after_failure.sh)** displays debugging information in case of failure.
```
TravisCarden, line 21: This is already done in #313. Let's remove the change here to avoid a merge conflict.

TravisCarden (changes requested):
This one's not ready yet. There are still quite a lot of references to Coveralls left. Be sure to search the codebase for the string "coveralls" in file contents and filenames. Don't forget to remove the `php-coveralls/php-coveralls` Composer library.

Also, we don't want to remove `after_script.sh`--it's used for telemetry. `after_success.sh` is the one that sends Coveralls data. Remove the Coveralls features from that file, but the file itself needs to remain, because users' configurations (i.e., GitHub Actions steps) reference it, and removing it would require all of our users to update their integrations or start having failures.

Note: you should rebase onto the current state of `develop`, while you're at it.

TravisCarden (approved):
_Approved._ Good-bye, Coveralls! 😄

PR #340, opened 2023-02-14
ORCA-486: Packages alter improvements

secretsayan:


TravisCarden:
@secretsayan, I pushed a commit to correct and expand test cases, which indicate a regression: an empty array is no longer treated as a valid package specification. That, of course, is a major use case. Please fix the problem and consider why no other tests caught this--where else we may need to expand tests to catch this kind of problem in the future.

TravisCarden (approved):

PR #343, opened 2023-02-21
ORCA-511: Add Functional Javascript Tests

secretsayan:
- We are adding a functional tests and a functional javascript test to ORCA's _example_ module so that during ORCA's self tests we can verify those and demonstrate ORCA's ability to support them.

- We are also printing out the command which is getting invoked while starting _chromedriver_ for functional javascript tests so as to keep parity with other scenarios where _chromedriver_ is used , for example Nightwatch.

- Use appropriate path when invoking _chromedriver_ for functional javascript tests.

### DIFF
--- a/example/tests/src/Functional/ExampleLoginTest.php
+++ b/example/tests/src/Functional/ExampleLoginTest.php
@@ -2,7 +2,6 @@
 
 namespace Drupal\Tests\example\Functional;
 
-// Use Drupal\FunctionalJavascriptTests\WebDriverTestBase;.
 use Drupal\Tests\BrowserTestBase;
 
 class ExampleLoginTest extends BrowserTestBase {
--- a/example/tests/src/Functional/ExampleLoginTest.php
+++ b/example/tests/src/Functional/ExampleLoginTest.php
@@ -1,17 +1,17 @@
 <?php
 
-namespace Drupal\Tests\example\FunctionalJavascriptTests;
+namespace Drupal\Tests\example\Functional;
 
 // Use Drupal\FunctionalJavascriptTests\WebDriverTestBase;.
 use Drupal\Tests\BrowserTestBase;
 
-class PageVisitTest extends BrowserTestBase {
+class ExampleLoginTest extends BrowserTestBase {
 
   protected $defaultTheme = 'claro';
 
   private $privilegedUser;
 
-  public function testPageVisit(): void {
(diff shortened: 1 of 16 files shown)


example/tests/src/Functional/ExampleLoginTest.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Drupal\Tests\example\Functional;
+
+// Use Drupal\FunctionalJavascriptTests\WebDriverTestBase;.
```
TravisCarden, line 5: What's this comment?

example/tests/src/FunctionalJavascriptTests/ExampleHomeLinkTest.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Drupal\Tests\example\FunctionalJavascriptTests;
```
TravisCarden, line 3: Based on the pattern in Drupal core, I think this namespace should be...

```suggestion
namespace Drupal\Tests\example\FunctionalJavascript;
```

(^^With the corresponding change to the directory name, obviously.)

src/Domain/Server/ServerBase.php
```diff
@@ -66,6 +66,8 @@ public function start(): void {
       ->disableOutput()
 
       ->start();
+
+    return $this->process->getCommandLine();
```
TravisCarden, line 70: Hmm. I don't like returning output here. Given how many data we _could_ return, it seems kind of arbitrary to pick just one. What if we returned the whole process object so client code could select anything it wanted? (Don't forget to update the docblock on the interface accordingly.)

src/Domain/Tool/TestRunner.php
```diff
@@ -182,7 +182,7 @@ public function setRunServers(bool $run_servers): void {
    */
   private function startServers(): void {
     $this->output->comment('Starting servers');
-    $this->serverStack->start();
+    $this->output->comment($this->serverStack->start());
```
TravisCarden, line 185: As discussed, we could make it clearer what this comment means:

```suggestion
    $this->output->comment('Running: ' . $this->serverStack->start());
```

TravisCarden (changes requested):
Looks great! A few changes needed.

TravisCarden (approved):
Looks great!

PR #347, opened 2023-03-13
ORCA-486: Fix packages alter issues

secretsayan:
Improve packages alter configuration parsing.


src/Domain/Package/PackageManager.php
```diff
@@ -181,7 +181,7 @@ private function initializePackages(FixturePathHandler $fixture_path_handler, st
       // Skipping a null datum provides for a package to be effectively removed
       // from the active specification at runtime by setting its value to NULL
       // in the packages configuration alter file.
-      if ($datum === NULL) {
+      if ($datum === NULL || (array_key_exists('core_matrix', $datum) && $datum['core_matrix'] === NULL)) {
```
TravisCarden, line 184: Wouldn't this (incorrectly) cause the following valid package to be removed? Better add a test case.

```yaml
drupal/example:
  core_matrix: ~
  version: 1.x
  version_dev: 1.x-dev
```

src/Domain/Package/PackageManager.php
```diff
@@ -190,19 +190,20 @@ private function initializePackages(FixturePathHandler $fixture_path_handler, st
...
       }
 
-      if ((array_key_exists('version', $datum) || array_key_exists('version_dev', $datum)) && $this->containsValidVersion($datum)) {
-        $this->addPackage($datum, $fixture_path_handler, $package_name);
-        continue;
-      }
-
-      if (isset($datum['core_matrix'])) {
+      if (array_key_exists('core_matrix', $datum)) {
         $constraints = array_values($datum['core_matrix']);
```
TravisCarden, line 194: What if `$datum['core_matrix']` isn't an array? Maybe we should throw an exception with an actionable error message.

src/Domain/Package/PackageManager.php
```diff
@@ -190,19 +190,20 @@ private function initializePackages(FixturePathHandler $fixture_path_handler, st
...
             $this->addPackage($datum, $fixture_path_handler, $package_name);
             break;
           }
         }
+        continue;
+      }
+
+      if (is_array($datum) && $this->containsValidVersion($datum)) {
+        $this->addPackage($datum, $fixture_path_handler, $package_name);
+        continue;
```
TravisCarden, line 206: I don't think this is necessary, is it? Since this is the last thing in the loop, it will continue either way, right?

src/Domain/Package/PackageManager.php
```diff
@@ -234,6 +235,11 @@ private function parseYamlFile(string $file): array {
    * Checks if a package is null.
    */
   private function containsValidVersion(array $data) : bool {
+
+    if (!array_key_exists('version', $data) && !array_key_exists('version_dev', $data)) {
```
TravisCarden, line 239: Do we require _both_ `version` _and_ `version_dev` to be present? If neither one is present, we currently just fall back to `*` for both, right? If only one is specified, should we just fall back to `*` on the other one?

Also, some comments about the motivation behind all these checks would be helpful.

tests/Domain/Package/PackageManagerTest.php
```diff
@@ -38,6 +38,7 @@ class PackageManagerTest extends TestCase {
         '*' => ['version' => '~1.0', 'version_dev' => '1.x-dev'],
       ],
     ],
+    'drupal/module6' => ['type' => 'composer-plugin'],
```
TravisCarden, line 41: Not to be pedantic, but if it's a Composer plugin, it can't be a Drupal module, right?

```suggestion
    'project/package' => ['type' => 'composer-plugin'],
```

src/Domain/Package/PackageManager.php
secretsayan: If `$datum['core_matrix']` isn't an array, the OptionsResolver is will throw an error pointing out that is should be an array and stops tests execution. Are you suggesting that we should show a message that this configuration is incorrect and remove the package from list of packages and continue the execution?

src/Domain/Package/PackageManager.php
secretsayan: Good catch :)

tests/Domain/Package/PackageManagerTest.php
secretsayan: Yes, correct.

src/Domain/Package/PackageManager.php
secretsayan: We cannot keep core_matrix as NULL as it will throw an error with the OptionsResolver where core_matrix is typed as an array.

src/Domain/Package/PackageManager.php
```diff
@@ -190,19 +190,20 @@ private function initializePackages(FixturePathHandler $fixture_path_handler, st
...
         foreach ($constraints as $constraint) {
           if ($this->containsValidVersion($constraint)) {
             $this->addPackage($datum, $fixture_path_handler, $package_name);
             break;
           }
         }
+        continue;
+      }
+
+      if (is_array($datum) && $this->containsValidVersion($datum)) {
```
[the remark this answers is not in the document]
secretsayan: Noted.

src/Domain/Package/PackageManager.php
TravisCarden: Oh, I see. Great; that's fine, then.

src/Domain/Package/PackageManager.php
TravisCarden: Ah. No, I'm satisfied that the `OptionsResolved` will throw an error. I didn't think of that.

src/Domain/Package/PackageManager.php
secretsayan: This condition is used to handle scenarios in which both `version` and `version_dev` are absent, like below -

`'drupal/add_me4' => ['type' => 'library']` .

I would put appropriate comments so that its easier to understand.

src/Domain/Package/PackageManager.php
```diff
@@ -234,6 +235,11 @@ private function parseYamlFile(string $file): array {
    * Checks if a package is null.
    */
   private function containsValidVersion(array $data) : bool {
+
+    if (!array_key_exists('version', $data) && !array_key_exists('version_dev', $data)) {
+      return TRUE;
+    }
+
     return (array_key_exists('version', $data) && !is_null($data['version'])) || (array_key_exists('version_dev', $data) && !is_null($data['version_dev']));
```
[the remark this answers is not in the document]
secretsayan: No, it handle scenarios which have something similar to below -

`'drupal/add_me4' => ['type' => 'library']`.

That means it returns `TRUE` for packages which do not have `version` and `version_dev` are both absent.

src/Domain/Package/PackageManager.php
```diff
@@ -190,19 +190,20 @@ private function initializePackages(FixturePathHandler $fixture_path_handler, st
...
         }
+        continue;
+      }
+
+      if (is_array($datum) && $this->containsValidVersion($datum)) {
+        $this->addPackage($datum, $fixture_path_handler, $package_name);
+        continue;
       }
 
     }
```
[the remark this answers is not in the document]
secretsayan: I have made some changes, please let me know if anything else is required to be added.

secretsayan:
@TravisCarden : Please review this again :)

src/Domain/Package/PackageManager.php
```diff
@@ -180,29 +195,41 @@ private function initializePackages(FixturePathHandler $fixture_path_handler, st
...
           if ($this->containsValidVersion($constraint)) {
             $this->addPackage($datum, $fixture_path_handler, $package_name);
             break;
           }
         }
+        continue;
+      }
+
+      // If all above conditions are not met we check to see if either of
+      // version and version_dev are defined.
```
TravisCarden, line 230: At this level of abstraction, all that matters is that the datum contains valid version information--which is already captured in the method name. We don't want to add distracting details here that are sure to get out of sync with the method logic over time.

```suggestion
```

TravisCarden (changes requested):

src/Domain/Package/PackageManager.php
```diff
@@ -180,29 +195,41 @@ private function initializePackages(FixturePathHandler $fixture_path_handler, st
...
         continue;
       }
 
-      if (is_array($datum) && count($datum) === 0) {
-        $this->addPackage($datum, $fixture_path_handler, $package_name);
+      // Remove packages which have declared core_matrix as something other than
+      // an array even if they declared a valid version as it will cause ORCA to
+      // throw an error in the Options resolver.
+      if ((array_key_exists('core_matrix', $datum) && !is_array($datum['core_matrix']))) {
+        $this->output->writeln('Skipping ' . $package_name . 'as core_matrix not defined as an array.');
```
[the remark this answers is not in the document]
secretsayan: I think as `OptionsResolver` is already handling this case and throwing an error appropriately we can remove the `core_matrix` type check here.

What do you suggest?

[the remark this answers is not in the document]
TravisCarden: I agree. 👍

TravisCarden (approved):
Great. Approved. 👍

PR #370, opened 2023-04-25
ORCA-525: Remove Ckeditor tests from acquia_dam

secretsayan:


src/Domain/Fixture/FixtureCustomizer.php
```diff
@@ -117,4 +118,48 @@ public function removePerzParagraphsTests(FixtureOptions $options): void {
...
+   *
+   * The package drupal/acquia_dam requires the ckeditor module for running
+   * its phpunit tests as it depends on classes only present in the ckeditor
+   * module. But ckeditor module is only a dev-dependency of
+   * drupal/acquia_dam and thus does not get included in the fixture unless
+   * drupal/acquia_dam is the SUT. This causes everyone else's builds to fail.
+   * So we are removing all tests requiring paragraphs module when
+   * drupal/acquia_dam is not the SUT.
+   */
+  public function removeAcquiaDamCkeditorTests(FixtureOptions $options): void {
```
TravisCarden, line 132: It looks like this method is identical to `:: removePerzParagraphsTests()`, except for the module name details. Why don't you extract a helper method you can call with just the values that differ between them? Something like this, for example:

```php
  public function removePerzParagraphsTests(FixtureOptions $options): void {
    $this->removeModuleDevDependencyTests($options, 'drupal/acquia_perz', 'Drupal\Tests\paragraphs');
  }

  public function removeAcquiaDamCkeditorTests(FixtureOptions $options): void {
    $this->removeModuleDevDependencyTests($options, 'drupal/acquia_dam', 'Drupal\Tests\ckeditor');
  }

  protected function removeModuleDevDependencyTests(
    FixtureOptions $options,
    string $module_name,
    string $search_string,
  ): void {
    // ...
  }
```

That raises the level of abstraction and makes the relevant details pop right out.

TravisCarden (changes requested):

src/Domain/Fixture/FixtureCustomizer.php
secretsayan: Done

secretsayan:
@TravisCarden : Please review.

src/Domain/Fixture/FixtureCustomizer.php
```diff
@@ -84,37 +85,79 @@ public function runCustomizations(FixtureOptions $options): void {
...
+
     $finder = $this->finderFactory->create();
+    // Converting drupal/acquia_dam to acquia_dam.
+    $module_name = explode("/", $module_name)[1];
 
     try {
-      $paragraph_files = $finder->in($this->fixturePathHandler
-        ->getPath('docroot/modules/contrib/acquia_perz'))
-        ->contains('Drupal\Tests\paragraphs');
+      $ckeditor_files = $finder->in($this->fixturePathHandler
```
TravisCarden, line 141: This variable needs to be renamed now.

secretsayan: Done

PR #379, opened 2023-05-05
Normalize example composer

secretsayan:
These changes are recommended by the latest version of `ergebnis/composer-normalize`.


secretsayan:
@TravisCarden : Merging this since we are just changing the order of already present key value pairs

PR #383, opened 2023-05-10
ORCA-532: Fix Previous Minor and Latest LTS

secretsayan:


src/Domain/Composer/Version/DrupalCoreVersionResolver.php
```diff
@@ -248,6 +253,14 @@ private function findPreviousMinor(): string {
     }
 
     $parts = explode('.', $this->findCurrent());
+
+    // If current minor version has '0' then
+    // previous minor version does not exist.
+    if ($parts[1] === '0') {
+      $message = "previous minor version not available";
+      throw new OrcaVersionNotFoundException($message);
+    }
```
TravisCarden, line 262: What would you think about this?

```suggestion
    // The "previous minor" version must be in the same major version. For
    // example, v10.1.0 has a previous minor version of v10.0.0. v10.0.0 has
    // no previous minor version. Therefore, if the current minor version is
    // zero (0), there is no previous minor.
    if ($parts[1] === '0') {
      $message = "There is no previous minor version available within the current major.";
      throw new OrcaVersionNotFoundException($message);
    }
```

tests/Domain/Ci/Job/IntegratedTestOnPreviousMinorCiJobTest.php
```diff
@@ -4,31 +4,38 @@
...
 use Acquia\Orca\Helper\EnvFacade;
 use Acquia\Orca\Helper\Process\ProcessRunner;
 use Acquia\Orca\Tests\Domain\Ci\Job\_Helper\CiJobTestBase;
 use Prophecy\Prophecy\ObjectProphecy;
+use Symfony\Component\Console\Output\OutputInterface;
 
 /**
- * @property \Acquia\Orca\Domain\Ci\Job\Helper\RedundantJobChecker|\Prophecy\Prophecy\ObjectProphecy $redundantJobChecker
+ * @property \Acquia\Orca\Domain\Ci\Job\Helper\RedundantJobChecker|\Prophecy\Prophecy\ObjectProphecy
+ *   $redundantJobChecker
```
TravisCarden, line 18: Does PHPCS accept this? I didn't think `@property` attributes were allowed to wrap.

tests/Domain/Composer/Version/DrupalCoreVersionResolverTest.php
```diff
@@ -15,19 +15,26 @@
...
- * @property \Composer\Package\PackageInterface|\Prophecy\Prophecy\ObjectProphecy $package
- * @property \Composer\Package\Version\VersionSelector|\Prophecy\Prophecy\ObjectProphecy $selector
+ * @property \Acquia\Orca\Domain\Composer\Version\DrupalDotOrgApiClient|\Prophecy\Prophecy\ObjectProphecy
+ *   $drupalDotOrgApiClient
+ * @property \Acquia\Orca\Domain\Composer\Version\VersionSelectorFactory|\Prophecy\Prophecy\ObjectProphecy
+ *   $selectorFactory
+ * @property \Composer\Package\PackageInterface|\Prophecy\Prophecy\ObjectProphecy
+ *   $package
+ * @property \Composer\Package\Version\VersionSelector|\Prophecy\Prophecy\ObjectProphecy
+ *   $selector
```
TravisCarden, line 25: Ditto above on wrapping `@property` annotations. Also, this sort of change is meaningless, i.e., it doesn't make any difference to the PR, so it just creates noise. It should be avoided.

TravisCarden (approved):
Looks good. If you make the few tiny changes I requested, you may go ahead and merge without re-review.

tests/Domain/Composer/Version/DrupalCoreVersionResolverTest.php
secretsayan: Done

tests/Domain/Ci/Job/IntegratedTestOnPreviousMinorCiJobTest.php
secretsayan: Fixed

PR #388, opened 2023-05-24
ORCA-543: Allow Integrated Next Minor jobs to fail

secretsayan:


secretsayan:
@TravisCarden Merging this.

PR #401, opened 2023-06-14
ORCA-527: Dependency management

secretsayan:


src/Domain/Fixture/SubextensionManager.php
```diff
@@ -95,7 +95,7 @@ public function __construct(ConfigLoader $config_loader, Filesystem $filesystem,
    *   The package manager.
    */
   public function initializeTopLevelExtensions(PackageManager $package_manager): void {
-    foreach ($package_manager->getAll() as $package_name => $package) {
+    foreach ($package_manager->getCompanyPackages() as $package_name => $package) {
```
TravisCarden, line 98: Could there ever be a reason to add a Drupal module as a dependency? What if AMA used this to install Paragraphs, for example? In any case, based solely on the `packages.yml` schema and documentation, I would expect it to apply to dependencies, too, if I were a user.

```suggestion
    foreach ($package_manager-> getAll() as $package_name => $package) {
```

src/Domain/Package/Package.php
```diff
@@ -74,6 +74,9 @@ class Package {
    *     both of the "version" and "version_dev" key-value pairs to be used when
    *     the corresponding Drupal core version constraint is satisfied. Mappings
    *     are processed in order, and the first match wins.
+   *   - "is_company_package": (optional) Whether package is a company package
+   *     or not. Defaults to "true" but can be overridden. Anything else is
+   *     always false and cannot be overridden.
```
TravisCarden, line 79: I notice this `@param` documentation has gotten out of sync with the actual `packages.yml` schema. (It's missing `enable`.) I wonder if we should just remove the details altogether and point to `packages.yml` for details. A follow-up PR would be fine for any changes like this if you feel like it would be cleaner. I would be happy for you to create one and merge it without review.

src/Domain/Package/PackageManager.php
```diff
@@ -130,6 +130,38 @@ public function getAll(): array {
...
+    return $companyPackages;
+  }
+
+  /**
+   * Gets an array of all dependencies.
+   *
+   * @return \Acquia\Orca\Domain\Package\Package[]
+   *   An array of packages keyed by package name.
+   */
+  public function getDependencies(): array {
```
TravisCarden, line 155: I wonder if we should make the distinction here a little more explicit, since company packages are technically dependencies, too. Like this, maybe?

```suggestion
  public function getThirdPartyDependencies(): array {
```

Or this?

```suggestion
  public function getVendorDependencies(): array {
```

TravisCarden (changes requested):
Looking good! A few questions/requests.

config/packages.yml
```diff
@@ -39,6 +39,9 @@
 # - "enable": Whether to enable the extension in Drupal. Defaults to true for
```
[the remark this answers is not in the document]
secretsayan: Done

src/Domain/Fixture/FixtureCreator.php
```diff
@@ -291,18 +278,6 @@ private function fixDefaultDependencies(): void {
 
     $additions[] = "drupal/core-dev:{$this->options->getCore()}";
 
-    if ($this->shouldRequireProphecyPhpunit()) {
-      $additions[] = 'phpspec/prophecy-phpunit:^2';
-    }
-
-    if ($this->shouldDowngradePhpunit()) {
-      $additions[] = 'phpunit/phpunit:9.4.3';
-    }
```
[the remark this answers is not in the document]
secretsayan: Let's move `phpspec/prophecy-phpunit:^2` to `packages.yml` . 

But the condition to down grade phpunit mentioned in the function `shouldDowngradePhpunit` will always be false. So this may be removed.

src/Domain/Fixture/FixtureCreator.php
```diff
@@ -316,49 +291,31 @@ private function fixDefaultDependencies(): void {
...
-   * @see https://www.drupal.org/project/drupal/issues/3186443
-   *
-   * @return bool
-   *   Returns TRUE if it should be downgraded, or FALSE if not.
+   * @throws \Acquia\Orca\Exception\OrcaException
    */
-  private function shouldDowngradePhpunit(): bool {
-    $version = $this->options->getCoreResolved();
-    return Comparator::equalTo($version, '9.1.0.0');
+  private function addDependencies(): array {
```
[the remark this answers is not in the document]
secretsayan: Yes, makes sense.

src/Domain/Fixture/FixtureInspector.php
```diff
@@ -331,7 +331,7 @@ private function getComposerLock(): Config {
    */
   private function getInstalledPackages(): array {
     $packages = [new TableSeparator()];
-    foreach ($this->packageManager->getAll() as $package_name => $package) {
+    foreach ($this->packageManager->getCompanyPackages() as $package_name => $package) {
```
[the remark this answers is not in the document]
secretsayan: I agree.

src/Domain/Fixture/SubextensionManager.php
secretsayan: Since we are already adding dependencies in the "fixDefaultDependencies" stage of fixture creation, would it make any difference? The dependency will be present in the fixture anyways.

src/Domain/Package/Package.php
secretsayan: Okay, I will create a PR for this.

src/Domain/Package/PackageManager.php
secretsayan: I will go with `getThirdPartyDependencies `

src/Domain/Package/Package.php
secretsayan: Created PR https://github.com/acquia/orca/pull/403 and merging.

src/Domain/Fixture/SubextensionManager.php
TravisCarden: Oh, I was mixed up about what code I was looking at here. Yes, this is fine. 🙂️

tests/Domain/Package/PackageManagerTest.php
```diff
@@ -142,9 +160,18 @@ public function testConstructionAndGetters(): void {
...
+    foreach (array_keys($all_dependencies) as $name) {
+      $actual_dependency_list[$name] = 0;
+    }
+
     self::assertEquals(self::EXPECTED_PACKAGE_LIST, $actual_package_list, 'Set/got all packages.');
     self::assertInstanceOf(Package::class, reset($all_packages), 'Got packages as Package objects.');
     self::assertEquals('drupal/module2', $package->getPackageName(), 'Got package by name.');
+    self::assertEquals(self::EXPECTED_DEPENDENCY_LIST, $actual_dependency_list, 'Set/got all dependencies.');
+    self::assertEquals(TRUE, $package->isCompanyPackage(), 'Set/got all dependencies.');
+    self::assertEquals(FALSE, $manager->get('drupal/dependency1')->isCompanyPackage(), 'Set/got all dependencies.');
```
TravisCarden, line 174: Looks like all three assertions accidentally got the same "Set/got all dependencies" assertion message.

TravisCarden (approved):
Looks great!

tests/Domain/Package/PackageManagerTest.php
secretsayan: oops, let me correct them.

PR #405, opened 2023-06-16
Fix exceptions

secretsayan:
- Since ORCA's self tests do not use the `config/packages.yml` we are adding the dependencies in `example/tests/packages_alter.yml` so that they can be picked up during fixture creation.

### DIFF
--- a/src/Helper/Log/GoogleApiClient.php
+++ b/src/Helper/Log/GoogleApiClient.php
@@ -212,7 +212,7 @@ class GoogleApiClient {
     }
     catch (ExceptionInterface $e) {
       $this->output->comment('An error occurred accessing the auth token from Google API endpoint.\n' . $e->getMessage());
-      exit;
+      return NULL;
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/Helper/Log/GoogleApiClient.php
```diff
@@ -212,7 +212,7 @@ public function getToken(): ?string {
     }
     catch (ExceptionInterface $e) {
       $this->output->comment('An error occurred accessing the auth token from Google API endpoint.\n' . $e->getMessage());
-      exit;
+      return NULL;
```
TravisCarden, line 215: You don't actually have to specify `return NULL`. Just `return` does the same thing.

```suggestion
      return;
```

TravisCarden (approved):
Looks good!

src/Helper/Log/GoogleApiClient.php
secretsayan: Its throwing error https://github.com/acquia/orca/actions/runs/5296402120/jobs/9587488078#step:5:1932

PR #406, opened 2023-06-16
ORCA-534: Add composer audit

secretsayan:


src/Domain/Fixture/FixtureCreator.php
```diff
@@ -212,6 +212,7 @@ public function create(FixtureOptions $options): void {
     $this->customizeFixture();
     $this->createAndCheckoutBackupTag();
     $this->displayStatus();
+    $this->auditPackages();
```
TravisCarden, line 215: Let's display the `audit` output before the status output, which I think makes the most sense at the end. And Let's change the method name to make it clear we're _displaying_ the audit results, not asserting them, or something--in other words, the result won't fail the build.

```suggestion
    $this->displaySecurityVulnerabilityAdvisories();
    $this->displayStatus();
```

src/Domain/Fixture/FixtureCreator.php
```diff
@@ -915,4 +916,17 @@ private function displayStatus(): void {
       ->render();
   }
 
+  /**
+   * Check for security vulnerability advisories for installed packages.
+   */
+  private function auditPackages(): void {
+    $this->output->section('Running Composer Audit');
```
TravisCarden, line 923: Let's make this more "intention revealing", i.e., saying what the user cares about, not what's happening under the hood (which they can see in the log output, if they really need to know). Maybe something like this:

```suggestion
    $this->output->section('Showing security vulnerability advisories');
```

TravisCarden (changes requested):
Looks good. I just have a couple of suggestions. Go ahead and merge if you address them.

PR #411, opened 2023-06-26
ORCA-531: Composer exit on patch failure

secretsayan:
Sets `composer-exit-on-patch-failure` to `false` for all the jobs which are allowed to fail.

### DIFF
--- a/src/Domain/Fixture/FixtureCreator.php
+++ b/src/Domain/Fixture/FixtureCreator.php
@@ -137,6 +137,8 @@ class FixtureCreator {
   private $fixtureCustomizer;
 
   /**
+   * The environment facade.
+   *
    * @var \Acquia\Orca\Helper\EnvFacade
    */
   private EnvFacade $env;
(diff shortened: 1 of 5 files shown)


src/Domain/Fixture/FixtureCreator.php
```diff
@@ -223,6 +251,27 @@ private function createComposerProject(): void {
...
+    $this->output->section("Setting composer-exit-on-patch-failure to false.");
+    try {
+      $this->composer->setConfig([
+        'extra.composer-exit-on-patch-failure',
+        'false',
+        '--json',
+      ]);
+    }
+    catch (\Exception $e) {
+      $this->output->writeln("Failed to remove Composer platform requirements.");
```
TravisCarden, line 271: This message looks like a copy/paste leftover. Maybe it's supposed to be something like this?

```suggestion
      $this->output->writeln("Failed to set composer-exit-on-patch-failure.");
```

TravisCarden (changes requested):
Just one little change and it can be merged.

PR #416, opened 2023-07-05
ORCA-551: Fix next minor dev calculations

secretsayan:
- If `next_minor_dev` exists then use it else make `next_major_latest_minor_dev` as `next_minor_dev` and skip running `next_major_latest_minor_dev` job


TravisCarden (approved):
Looks good to me!

PR #420, opened 2023-08-03
Orca 559 : Compatibility with PHPunit 10

secretsayan:
- Coverage was enabled for all tests, as tests were failing due to a warning saying "coverage driver is not available" , may be we should resolve "failonwarning" settings of phpunit


tests/Domain/Composer/Version/DrupalCoreVersionResolverTest.php
```diff
@@ -446,8 +446,9 @@ public function testResolvePredefinedVersionNotFound($version): void {
     $resolver->resolvePredefined($version);
   }
 
-  public function providerResolvePredefinedVersionNotFound(): array {
-    $data = $this->providerVersions();
+  public static function providerResolvePredefinedVersionNotFound(): array {
+    $obj = new DrupalCoreVersionResolverTest('testResolvePredefinedVersionNotFound');
+    $data = $obj->providerVersions();
```
TravisCarden, line 451: I think you can still use the original function if you just call it statically:

```suggestion
    $data = self::providerVersions();
```

tests/Options/CiRunOptionsTest.php
```diff
@@ -106,25 +106,26 @@ public function testMissingRequiredOptions($options): void {
...
+  public static function providerMissingRequiredOptions(): array {
+    $obj = new CiRunOptionsTest("testMissingRequiredOptions");
     return [
       'No options' => [[]],
       'Missing job' => [
         [
-          'phase' => $this->validPhaseName(),
-          'sut' => $this->validSutName(),
+          'phase' => $obj->validPhaseName(),
+          'sut' => $obj->validSutName(),
```
TravisCarden, line 116: I would just make all these helper methods static on `CiJobTestBase` instead of duplicating their logic.

.github/workflows/orca.yml
```diff
@@ -66,7 +66,7 @@ jobs:
           - INTEGRATED_TEST_ON_NEXT_MAJOR_LATEST_MINOR_DEV
         php-version: [ "8.1", "8.3" ]
         orca-enable-nightwatch: [ "FALSE" ]
-        orca-coverage-enable: [ "FALSE" ]
+        orca-coverage-enable: [ "TRUE" ]
```
TravisCarden, line 69: You probably told me why we're doing this, but I just want to call it out in case it was unintentional.

TravisCarden (changes requested):
Looks good in general, with a few suggestions. (I hope there was a script to change all those data provider signatures to static for you!)

As to the removal of PHPLOC, that's a functional change that should have a separate PR. No one looking at this PR would say, "Hmm, Compatibility with PHPunit 10. I'd better make sure it didn't change the CLI API." Also, when you announce the change, you'll want to link to a PR where it and it only is discussed.

phpunit.xml.dist.bak
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
secretsayan: Yes..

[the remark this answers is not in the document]
secretsayan: Removed now

tests/Options/CiRunOptionsTest.php
secretsayan: We are using all this helper functions from the `CiEnumsTestTrait`, do you mean we need to make them `static` there?

.github/workflows/orca.yml
secretsayan: I will leave a comment in the `orca.yml` file mentioning this.

tests/Options/CiRunOptionsTest.php
TravisCarden: Yes, let's make them static anywhere they appear.

src/Console/Command/Qa/QaStaticAnalysisCommand.php
```diff
@@ -147,14 +137,13 @@ protected function configure():void {
...
       ->addOption('composer', NULL, InputOption::VALUE_NONE, 'Run the Composer validation tool')
-      ->addOption('coverage', NULL, InputOption::VALUE_NONE, 'Run the code coverage estimator. Implies "--phploc"')
+      ->addOption('coverage', NULL, InputOption::VALUE_NONE, 'Run the code coverage estimator.')
       ->addOption('phpcs', NULL, InputOption::VALUE_NONE, 'Run the PHP Code Sniffer tool')
       ->addOption('phpcs-standard', NULL, InputOption::VALUE_REQUIRED, implode(PHP_EOL, array_merge(
         ['Change the PHPCS standard used:'],
         PhpcsStandardEnum::commandHelp()
       )), $this->defaultPhpcsStandard)
       ->addOption('phplint', NULL, InputOption::VALUE_NONE, 'Run the PHP Lint tool')
-      ->addOption('phploc', NULL, InputOption::VALUE_NONE, 'Run the PHPLOC tool')
```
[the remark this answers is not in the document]
secretsayan: This has been handled in https://github.com/acquia/orca/pull/528.

secretsayan:
@TravisCarden : Please review this PR, I have deprected `phploc` in https://github.com/acquia/orca/pull/528 and rebased with latest changes of develop.

src/Console/Command/Qa/QaStaticAnalysisCommand.php
```diff
@@ -148,7 +148,7 @@ protected function configure():void {
       ->setHelp('Tools can be specified individually or in combination. If none are specified, all will be run.')
       ->addArgument('path', InputArgument::REQUIRED, 'The path to analyze')
       ->addOption('composer', NULL, InputOption::VALUE_NONE, 'Run the Composer validation tool')
-      ->addOption('coverage', NULL, InputOption::VALUE_NONE, 'Run the code coverage estimator. Implies "--phploc"')
+      ->addOption('coverage', NULL, InputOption::VALUE_NONE, 'Run the code coverage estimator.')
```
TravisCarden, line 151: Ooh... I forgot about this. 😬 The health score calculator depends on PHPLOC data. Okay, we need to take this to the stakeholders.

secretsayan:
@TravisCarden  Please review this PR

.phpunit.result.cache
TravisCarden, line 1: This is an ephemeral file that needs to be git-ignored.

src/Console/Command/Qa/QaStaticAnalysisCommand.php
TravisCarden: This change should be reverted. It's accurate again now that PHPLOC is remaining.

TravisCarden (changes requested):
I'm gonna start being a little rougher on my reviews, @secretsayan. 😉 Please take this one back and look it over again before I review it. After a quick skim, I see some fairly basic oversights.

composer.json
```diff
@@ -59,7 +59,9 @@
         "brainmaestro/composer-git-hooks": "^2.8.5",
         "phpspec/prophecy-phpunit": "^2.0.2",
         "phpstan/phpstan": "^1.10.14",
-        "phpunit/phpunit": "^9.6.7"
+        "phpunit/php-code-coverage": "^10",
+        "phpunit/php-file-iterator": "^4.0",
```
TravisCarden, line 63: I forget: Is this here because we had to pin to a specific version that would be installable on all supported versions of PHP? Because there's no way we're actually using it directly. If so, cool. If not, let's remove it.

TravisCarden (approved):
Looks good, @secretsayan! I left one question that you can mark resolved or address based on the answer. Merge at your pleasure. 🙂

composer.json
secretsayan: We are using this in our static analysis tests , check here https://github.com/acquia/orca/actions/runs/9027070083/job/24805359219?pr=420#step:9:161. Marking this as resolved and merging

PR #424, opened 2023-08-22
ORCA-568: Ignore push builds for dependabot

danepowell:
Currently when `dependabot` pushes a commit in a pull request it triggers two jobs :
1. Firstly, the job for the `pull request changes` happening due to this commit.
2. Secondly, the job for the `push event` in the pull request branch created in origin by `dependabot`

This PR intends to eliminate the second one i.e. triggering of a job due to a `push event` in a branch created by `dependabot` as it is redundant.

### DIFF
--- a/example/.github/workflows/orca.yml
+++ b/example/.github/workflows/orca.yml
@@ -24,7 +24,10 @@
 ---
 name: ORCA CI
 on:
-  push: ~
+  push:
+    # Prevent duplicate jobs on Dependabot PRs that interfere with automerge.
+    branches-ignore:
+      - 'dependabot/**'
   pull_request: ~
   schedule:
     # Daily at 00:00:00 UTC.


TravisCarden:
@danepowell, with this change, what would happen if Dependabot rebased a pull request? It seems to me that the updated pull request would not trigger a build. Does that sound right?

danepowell:
A rebase shouldn't be different from any other commit to a pull request in terms of workflow events. It will still trigger a new workflow run. We've been using this on Acquia CLI for over a year and Dependabot still happily does its thing: https://github.com/acquia/cli/blob/main/.github/workflows/ci.yml#L6

TravisCarden:
Okay, so just to make sure I understand you correctly, @danepowell, with this change, Dependabot PRs _and rebases done by Dependabot_ will both trigger a build? But if a user pushes to a Dependabot PR it _won't_ trigger a build? If I understand properly, the only behavior this would change is that if a human user commits and pushes up a new commit, the CI job won't run and the new commit won't be tested. Why is that desirable?

TravisCarden:
> This is due to "pushes"... and "pull request changes" being two separate events that aren't mutually exclusive.

_Okay_, that's key. In that case, this makes sense to me. @secretsayan, I would encourage you to add a brief inline code comment summarizing this discussion.

danepowell:
@secretsayan I intended to change the example file since this is what downstream projects (such as my own) will reference.

If _you_ want to update the root orca.yml, that's up to you. Let me know if you want me to add that to this PR.

secretsayan:
@danepowell thanks for the information. I will add it to the root in a different PR.

PR #438, opened 2023-11-08
ORCA-561: Domo Improvements

secretsayan:
- This adds "AllowedToFail" information in Domo charts so that we can clearly segregate tests that are actual failures and those which are allowed failures.
- This adds "Branch" information in Domo charts so that SUT branch is reported in dashboard.
- Run NIghtwatch tests before other Phpunit tests so that Nightwatch tests results can be posted in Domo.

### DIFF
--- a/bin/ci/_includes.sh
+++ b/bin/ci/_includes.sh
@@ -122,7 +122,6 @@ allowed_failures=(
   "INTEGRATED_TEST_ON_NEXT_MAJOR_LATEST_MINOR_BETA_OR_LATER"
 )
 if [[ " ${allowed_failures[*]} " =~ " ${ORCA_JOB} " ]]; then
-#  set +e
   export ORCA_IS_ALLOWED_FAILURE="TRUE"
   notice "This job is allowed to fail and will report as passing regardless of outcome."
 fi
--- a/bin/ci/_includes.sh
+++ b/bin/ci/_includes.sh
@@ -122,10 +122,18 @@ allowed_failures=(
   "INTEGRATED_TEST_ON_NEXT_MAJOR_LATEST_MINOR_BETA_OR_LATER"
 )
 if [[ " ${allowed_failures[*]} " =~ " ${ORCA_JOB} " ]]; then
-  set +e
+#  set +e
   export ORCA_IS_ALLOWED_FAILURE="TRUE"
   notice "This job is allowed to fail and will report as passing regardless of outcome."
 fi
+
+function shutdown() {
+    if [[ "$ORCA_IS_ALLOWED_FAILURE" == "TRUE" ]]; then
+        exit 0
+    fi
+}
+
+trap shutdown ERR
 
 # Make the shell print all lines in the script before executing them.
 set -v
--- a/bin/ci/script.sh
+++ b/bin/ci/script.sh
@@ -37,4 +37,3 @@ if [[ "$ORCA_JOB" ]]; then
   eval "orca ci:run $ORCA_JOB script $ORCA_SUT_NAME"
 fi
 
-echo "completed"
--- a/bin/ci/script.sh
+++ b/bin/ci/script.sh
@@ -37,4 +37,4 @@ if [[ "$ORCA_JOB" ]]; then
   eval "orca ci:run $ORCA_JOB script $ORCA_SUT_NAME"
 fi
 
-
+echo "completed"
--- a/bin/ci/script.sh
+++ b/bin/ci/script.sh
@@ -32,8 +32,9 @@ if [[ "$ORCA_ENABLE_NIGHTWATCH" == "TRUE" && "$ORCA_SUT_HAS_NIGHTWATCH_TESTS" &&
     kill -0 $CHROMEDRIVER_PID
   )
 fi
-
+echo $?
 if [[ "$ORCA_JOB" ]]; then
   eval "orca ci:run $ORCA_JOB script $ORCA_SUT_NAME"
+  echo $?
 fi
 
--- a/bin/ci/script.sh
+++ b/bin/ci/script.sh
@@ -12,7 +12,7 @@
 cd "$(dirname "$0")" || exit; source _includes.sh
 
 function shutdown() {
-    if (ORCA_IS_ALLOWED_FAILURE === TRUE); then
+    if ($ORCA_IS_ALLOWED_FAILURE === TRUE); then
       return 0;
     fi
 }
--- a/bin/ci/script.sh
+++ b/bin/ci/script.sh
@@ -36,3 +36,5 @@ fi
 if [[ "$ORCA_JOB" ]]; then
   eval "orca ci:run $ORCA_JOB script $ORCA_SUT_NAME"
 fi
+
+
--- a/bin/ci/script.sh
+++ b/bin/ci/script.sh
@@ -11,7 +11,12 @@
 
 cd "$(dirname "$0")" || exit; source _includes.sh
 
+echo "Debug 1: "$?
+if [[ "$ORCA_JOB" ]]; then
+  eval "orca ci:run $ORCA_JOB script $ORCA_SUT_NAME"
 
+fi
+echo "Debug 2: "$?
 
 if [[ "$ORCA_ENABLE_NIGHTWATCH" == "TRUE" && "$ORCA_SUT_HAS_NIGHTWATCH_TESTS" && -d "$ORCA_YARN_DIR" ]]; then
   (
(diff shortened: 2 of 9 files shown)


secretsayan:
@TravisCarden Need your review on this one

src/Console/Command/Ci/CiRunCommand.php
```diff
@@ -131,16 +142,24 @@ public function execute(InputInterface $input, OutputInterface $output): int {
...
       $output->writeln("Error: {$e->getMessage()}");
+      if ($this->env->get("ORCA_IS_ALLOWED_FAILURE")) {
+        return StatusCodeEnum::OK;
+      }
       return StatusCodeEnum::ERROR;
     }
     catch (\Exception $e) {
       $event = new CiEvent($data);
       $this->eventDispatcher->dispatch($event, CiEvent::NAME);
+      if ($this->env->get("ORCA_IS_ALLOWED_FAILURE")) {
```
secretsayan, line 160: @TravisCarden : Do you think this would be problematic?

src/Console/Command/Ci/CiRunCommand.php
```diff
@@ -131,16 +142,24 @@ public function execute(InputInterface $input, OutputInterface $output): int {
...
+      $data['allowedToFail'] = $this->env->get("ORCA_IS_ALLOWED_FAILURE") ?? FALSE;
+      $data['branch'] = $this->env->get("ORCA_SUT_BRANCH") ?? FALSE;
       $job->run($options);
 
     }
     catch (OrcaInvalidArgumentException $e) {
       $output->writeln("Error: {$e->getMessage()}");
+      if ($this->env->get("ORCA_IS_ALLOWED_FAILURE")) {
+        return StatusCodeEnum::OK;
+      }
```
secretsayan, line 154: @TravisCarden : Do you think this would be problematic?

src/Console/Command/Ci/CiRunCommand.php
```diff
@@ -131,16 +142,24 @@ public function execute(InputInterface $input, OutputInterface $output): int {
       $data['status'] = 'FAIL';
       $job = $this->ciJobFactory->create($options->getJob());
       $data['version'] = $job->getDrupalCoreVersion();
+      $data['allowedToFail'] = $this->env->get("ORCA_IS_ALLOWED_FAILURE") ?? FALSE;
+      $data['branch'] = $this->env->get("ORCA_SUT_BRANCH") ?? FALSE;
```
TravisCarden, line 146: I think these could just use the `$default` argument. Right?

```suggestion
      $data['allowedToFail'] = $this->env->get("ORCA_IS_ALLOWED_FAILURE", FALSE);
      $data['branch'] = $this->env->get("ORCA_SUT_BRANCH", FALSE);
```

https://github.com/acquia/orca/blob/adf0f9b0867ac389724d2e2ab4e0508082545028/src/Helper/EnvFacade.php#L12-L23

^ Incidentally, it might be helpful to expand the `@param` comment to clarify what "default return value" means, i.e., the value to return if the environment variable isn't set.

src/Console/Command/Ci/CiRunCommand.php
TravisCarden: No, I think I would let this one return an error code, because it's not a test failure that should be allowable--it means that something broke with the setup that needs to be surfaced and fixed.

src/Console/Command/Ci/CiRunCommand.php
TravisCarden: I think this is the same as the other one: https://github.com/acquia/orca/pull/438/files#r1388318264.

src/Helper/Log/GoogleApiClient.php
```diff
@@ -64,7 +64,7 @@ class GoogleApiClient {
   /**
    * The sheet id of the spreadsheet.
    */
-  private const SHEET_ID = "Sheet1";
+  private const SHEET_ID = "Sheet2";
```
TravisCarden, line 67: Is this a deliberate change? (It seems like it could be left over from testing and debugging.) Maybe the tests should cover this detail.

bin/ci/script.sh
```diff
@@ -11,9 +11,7 @@
 
 cd "$(dirname "$0")" || exit; source _includes.sh
 
-if [[ "$ORCA_JOB" ]]; then
-  eval "orca ci:run $ORCA_JOB script $ORCA_SUT_NAME"
-fi
+
```
TravisCarden, line 14: Just double-checking: was this a deliberate change to move this? Doesn't seem like it would make much difference in the end.

tests/Console/Command/Ci/CiRunCommandTest.php
```diff
@@ -54,13 +58,18 @@ protected function setUp(): void {
     $this->eventDispatcher
       ->dispatch(Argument::cetera())
       ->willReturn(new \stdClass());
+    $this->env = $this->prophesize(EnvFacade::class);
+    $this->env
+      ->get(Argument::any())
+      ->willReturn();
```
TravisCarden, line 64: I didn't know you could call `::willReturn()` without any arguments. I assume that will just make it return `NULL`. It might be good to be explicit for readability:

```suggestion
      ->willReturn(NULL);
```

tests/Helper/Log/GoogleApiClientTest.php
```diff
@@ -99,6 +101,8 @@ public function testPostDataWithLatestDrupalVersion(): void {
       'sut' => 'drupal/example',
       'status' => 'PASS',
       'version' => DrupalCoreVersionEnum::CURRENT(),
+      'allowedToFail' => TRUE,
```
TravisCarden, line 104: Should there be a test case somewhere for `'allowedToFail' => FALSE`?

TravisCarden (changes requested):

tests/Helper/Log/GoogleApiClientTest.php
secretsayan: No, I don't think so, we are just testing data transfer here, what value we are sending that won't impact any parts of ORCA.

src/Helper/Log/GoogleApiClient.php
secretsayan: Good catch

src/Console/Command/Ci/CiRunCommand.php
secretsayan: Okay

src/Console/Command/Ci/CiRunCommand.php
secretsayan: okay

bin/ci/script.sh
```diff
@@ -34,3 +32,9 @@ if [[ "$ORCA_ENABLE_NIGHTWATCH" == "TRUE" && "$ORCA_SUT_HAS_NIGHTWATCH_TESTS" &&
     kill -0 $CHROMEDRIVER_PID
   )
 fi
+
+if [[ "$ORCA_JOB" ]]; then
+  eval "orca ci:run $ORCA_JOB script $ORCA_SUT_NAME"
+fi
+
+echo "completed"
```
secretsayan, line 40: @TravisCarden : I am observing a weird behaviour, though this line has no significance at all, if I remove this line the allowed failure feature is not working and tests are failing. Do you know why?

TravisCarden: That _is_ weird; I _don't_ know why. Would it behave the same way with lines 36-38 back up at the beginning of the file where they were before? Because presumably this was working before, and that's the only significant difference I see.

secretsayan: If we put 36-38 where it was before, it would work absolutely fine. But I wanted to run the `Nightwatch` tests before we run `orca ci:run $ORCA_JOB script $ORCA_SUT_NAME` so that we can capture Nightwatch failures in Domo. But if I re-arrange the lines then the `allowed failure` feature is not working not sure why. Then I have to add the line `echo "completed"` or anything else to make it work. 

In short, it seems we cannot put the `orca ci:run $ORCA_JOB script $ORCA_SUT_NAME` part at the bottom of the file.

TravisCarden: Well, then I'd say choose which is more important to you (behavior that your fully understand vs. behavior that does what you want) and leave inline comments. If you're dissatisfied with that, take a personal note--or if it's important enough to you, create a Jira issue--and come back to it when it offers more value than competing priorities. `¯\_(ツ)_/¯`

secretsayan:
@TravisCarden : Please review again and let me know if any changes are required.

bin/ci/_includes.sh
```diff
@@ -122,10 +122,18 @@ allowed_failures=(
   "INTEGRATED_TEST_ON_NEXT_MAJOR_LATEST_MINOR_BETA_OR_LATER"
 )
 if [[ " ${allowed_failures[*]} " =~ " ${ORCA_JOB} " ]]; then
-  set +e
+#  set +e
```
TravisCarden, line 125: There's no reason to keep commented out code. If it's not worth a comment explaining why it's there, it's not worth keeping.

TravisCarden, line 125: There's no reason to keep commented out code. If it's not worth a comment explaining why it's there, it's not worth keeping.

TravisCarden (approved):
Nice! I have one trivial comment, and then I approve. No need to request another review for it.

bin/ci/_includes.sh
secretsayan: Done

PR #474, opened 2023-12-20
Modify DRUPAL_NIGHTWATCH_SEARCH_DIRECTORY variable to discover SUT

secretsayan:
After Drupal 10.2.0 release we are observing that Drupal is not detecting Nightwatch tests contained in the SUT. This happens because the SUT present in the `contrib` folder is only symlink to the actual SUT directory which is present adjacent to ORCA fixture directory. This PR tells Drupal to look for Nighwatch tests present in the actual SUT directory.


TravisCarden (approved):
Nice. 👍

secretsayan:
Thanks @TravisCarden.

@effulgentsia : Do you have any suggestions/comments or shall we go ahead and merge this PR?

PR #483, opened 2024-01-09
Changes related to accommodate nightwatch tests in Gitlab CI

secretsayan:
There are three changes - 

- Nightwatch and Functional Javascript : We have made changes in this PR so that we can use a `separate container` like `drupal.org` templates.
- For tests with Coverage : We are generating report using `Cobertura` format instead of `Clover`. Other changes related to names of variable from `clover` to `cobertura` will be handled in https://github.com/acquia/orca/pull/484 and not here due to time constraints.
- We are generating `junit` log for all the `phpunit` tests so that all tests and reports are shown in the Tests tab of Gitlab CI.

PR #484, opened 2024-01-18
Change coverage clover to cobertura for Gitlab CI

secretsayan:
Related to https://github.com/acquia/orca/pull/483

The important change which would impact cover generation is already merged here https://github.com/acquia/orca/pull/483/commits/9e2000fffb169e5c149fedbe20ce32bf4fb48fc6 , this PR deals with mainly the variable names and text.


secretsayan:
@TravisCarden : Please review the changes.

src/Enum/EnvVarEnum.php
```diff
@@ -102,7 +102,7 @@ class EnvVarEnum extends Enum {
    */
   public static function descriptions(): array {
     return [
-      self::ORCA_COVERAGE_CLOVER => 'The path where ORCA saves the PHPUnit test coverage Clover XML file',
+      self::ORCA_COVERAGE_COBERTURA => 'The path where ORCA saves the PHPUnit test coverage Clover XML file',
```
TravisCarden, line 105:
```suggestion
      self::ORCA_COVERAGE_COBERTURA => 'The path where ORCA saves the PHPUnit test coverage Cobertura XML file',
```

TravisCarden (commented):
Looks pretty good, @secretsayan. Are we sure no one is using the current clover files? I seem to remember that we first introduced the `$ORCA_COVERAGE_CLOVER` environment variable at Matt Glaman's request. Please check with (at least) him to make sure we don't break anything he's depending on.

secretsayan:
> Looks pretty good, @secretsayan. Are we sure no one is using the current clover files? I seem to remember that we first introduced the `$ORCA_COVERAGE_CLOVER` environment variable at Matt Glaman's request. Please check with (at least) him to make sure we don't break anything he's depending on.

Yeah, now that I think of it, I think you are right, multiple projects are using ORCA_COVERAGE_CLOVER variable to specify the path to a clover file to generate coverage and it would be a breaking change for all those projects. Instead, we can keep both the variables ORCA_COVERAGE_CLOVER and ORCA_COVERAGE_COBERTURA so that ORCA users can define and use whichever they like. What to do you think?

docs/advanced-usage.md
```diff
@@ -7,10 +7,10 @@ Various aspects of ORCA's behavior can be altered at runtime through the use of
 ### Command-line application
 
 These affect ORCA in all contexts.
-
```
TravisCarden, line 10: Let's leave this line break in here for consistency with the rest of the file.

docs/advanced-usage.md
```diff
@@ -7,10 +7,10 @@ Various aspects of ORCA's behavior can be altered at runtime through the use of
...
 * <a name="ORCA_COVERAGE_CLOVER"></a>**`ORCA_COVERAGE_CLOVER`**: Change the path where ORCA saves the PHPUnit test coverage Clover XML file.
+* <a name="ORCA_COVERAGE_COBERTURA"></a>**`ORCA_COVERAGE_COBERTURA`**: Change the path where ORCA saves the PHPUnit test coverage Clover XML file.
 
-* <a name="ORCA_COVERAGE_ENABLE"></a>**`ORCA_COVERAGE_ENABLE`**: Set to `TRUE` to generate test coverage data . The resulting Clover file will be saved at the location set in [`ORCA_COVERAGE_CLOVER`](#ORCA_COVERAGE_CLOVER). Test coverage generation greatly increases build times, so only enable it on one job--all that makes sense anyway.
+* <a name="ORCA_COVERAGE_ENABLE"></a>**`ORCA_COVERAGE_ENABLE`**: Set to `TRUE` to generate test coverage data . The resulting Clover/Cobertura file will be saved at the location set in [`ORCA_COVERAGE_COBERTURA`](#ORCA_COVERAGE_COBERTURA) or [`ORCA_COVERAGE_CLOVER`](#ORCA_COVERAGE_CLOVER) based on the config used. Test coverage generation greatly increases build times, so only enable it on one job--all that makes sense anyway.
```
TravisCarden, line 13: This doesn't make it quite clear whether you can have _both_ Clover _and_ Cobertura. Let's clarify.

src/Domain/Tool/Phpunit/PhpUnitTask.php
```diff
@@ -324,7 +324,7 @@ protected function runPhpUnit(): void {
         '--verbose',
       ];
       if ($this->shouldGenerateCodeCoverage()) {
-        $command[] = "--coverage-cobertura={$this->cloverCoverage}";
+        $command[] = "--coverage-cobertura={$this->coberturaCoverage}";
```
TravisCarden, line 327: We still need `--coverage-cobertura` _and_ `--coverage-clover`.  Can we generate both on the same run? If not, we'll need to determine which to generate based on environment variables.

secretsayan: Yeah, I will create separate environment variables like by default it will generate coverage in CLOVER but if `ORCA_COVERAGE_COBERTURA_ENABLE` is set then it will generate coverage in COBERTURA.

docs/advanced-usage.md
secretsayan: Updated text.

src/Domain/Tool/Phpunit/PhpUnitTask.php
secretsayan: Done.

secretsayan:
@TravisCarden : I re-wrote the PR and added additional variables, please review when you can.

docs/advanced-usage.md
```diff
@@ -9,8 +9,11 @@ Various aspects of ORCA's behavior can be altered at runtime through the use of
 These affect ORCA in all contexts.
 
 * <a name="ORCA_COVERAGE_CLOVER"></a>**`ORCA_COVERAGE_CLOVER`**: Change the path where ORCA saves the PHPUnit test coverage Clover XML file.
+* <a name="ORCA_COVERAGE_COBERTURA"></a>**`ORCA_COVERAGE_COBERTURA`**: Change the path where ORCA saves the PHPUnit test coverage Cobertura XML file.
+* <a name="ORCA_COVERAGE_COBERTURA_ENABLE"></a>**`ORCA_COVERAGE_COBERTURA_ENABLE`**: Set to `TRUE` to generate test coverage data in Cobertura format instead of the default Clover format.
 
-* <a name="ORCA_COVERAGE_ENABLE"></a>**`ORCA_COVERAGE_ENABLE`**: Set to `TRUE` to generate test coverage data . The resulting Clover file will be saved at the location set in [`ORCA_COVERAGE_CLOVER`](#ORCA_COVERAGE_CLOVER). Test coverage generation greatly increases build times, so only enable it on one job--all that makes sense anyway.
+* <a name="ORCA_COVERAGE_ENABLE"></a>**`ORCA_COVERAGE_ENABLE`**: Set to `TRUE` to generate test coverage data . By default, the coverage will be generated in Clover format and resulting Clover file will be saved at the location
```
TravisCarden, line 15: I'm afraid we've lost some symmetry here--we have one variable that enables two things while the other enables only one, and one feature that can't be disabled without disabling both. I think what we want is this:

| Enable | Configure |
|---|---|
| `ORCA_COVERAGE_CLOVER_ENABLE` | `ORCA_COVERAGE_CLOVER` |
| `ORCA_COVERAGE_COBERTURA_ENABLE` | `ORCA_COVERAGE_COBERTURA` |
| `ORCA_COVERAGE_ENABLE`* | `ORCA_COVERAGE_CLOVER` |

*For backward compatibility, `ORCA_COVERAGE_ENABLE` should be preserved with the same behavior, thus becoming essentially an alias of the new `ORCA_COVERAGE_CLOVER_ENABLE`. It should _not_ automatically enable Cobertura coverage.

TravisCarden (changes requested):
Looks good overall, but I think we need to make one change:

secretsayan:
@TravisCarden : Ready for review.

src/Domain/Tool/Phpunit/PhpUnitTask.php
```diff
@@ -323,10 +381,17 @@ protected function runPhpUnit(): void {
...
+
+      if ($this->shouldGenerateCodeCoverageInClover()) {
+        $command[] = "--coverage-clover={$this->cloverCoverage}";
+      }
+
+      if ($this->shouldGenerateCodeCoverageInCobertura()) {
+        $command[] = "--coverage-cobertura={$this->coberturaCoverage}";
       }
+
+      $this->processRunner->addEnvVar("XDEBUG_MODE", "coverage");
```
TravisCarden, line 393: So this environment variable was previously being set only if Clover coverage was enabled. Now it's being set under all circumstances--including if neither Clover _nor_ Cobertura is being generated. Was that a conscious decision?

bin/ci/self-test/script.sh
```diff
@@ -15,13 +15,13 @@ cd ../../../ || exit 1
...
+  echo "Coverage generation is on but Xdebug is disabled"
   exit 1
 fi
 
-if [[ "$ORCA_COVERAGE_ENABLE" == FALSE && "$XDEBUG_IS_ENABLED" == "TRUE" ]]; then
-  echo "ORCA_COVERAGE_ENABLE is off but Xdebug is enabled"
+if [[ "$ORCA_COVERAGE_ENABLE" == FALSE  && "$ORCA_COVERAGE_COBERTURA_ENABLE" == FALSE  && "$ORCA_COVERAGE_CLOVER_ENABLE" == FALSE ]] && [ "$XDEBUG_IS_ENABLED" == TRUE ]; then
+  echo "Coverage generation is off but Xdebug is enabled"
   exit 1
 fi
```
TravisCarden, line 26: These conditions are complex and hard to read. They suggest a need for refactoring and simplification. What if somewhere in the includes we automatically set `$ORCA_COVERAGE_ENABLE` to `TRUE` if _either_ `$ORCA_COVERAGE_COBERTURA_ENABLE` or `$ORCA_COVERAGE_CLOVER_ENABLE` is on and then use it under the hood for simpler comparisons. In fact, if we did that, I don't think these conditional tests would have to change at all. (Later, since we're deprecating `$ORCA_COVERAGE_ENABLE`, we can continue to use it under the hood and just not allow users to override it.)

src/Enum/EnvVarEnum.php
```diff
@@ -38,8 +39,15 @@ class EnvVarEnum extends Enum {
...
 
+  public const ORCA_COVERAGE_COBERTURA = 'ORCA_COVERAGE_COBERTURA';
+
+  public const ORCA_COVERAGE_COBERTURA_ENABLE = 'ORCA_COVERAGE_COBERTURA_ENABLE';
+
+  public const ORCA_COVERAGE_CLOVER_ENABLE = 'ORCA_COVERAGE_CLOVER_ENABLE';
+
   public const ORCA_COVERAGE_ENABLE = 'ORCA_COVERAGE_ENABLE';
 
+
```
TravisCarden, line 50:
```suggestion
```

TravisCarden (changes requested):

bin/ci/self-test/script.sh
secretsayan: I think it would create a conflict if we automatically set $`ORCA_COVERAGE_ENABLE` to `TRUE` when `ORCA_COVERAGE_COBERTURA_ENABLE` is _ON_ because `ORCA_COVERAGE_ENABLE` is an alias of `$ORCA_COVERAGE_CLOVER_ENABLE`.

.github/workflows/orca.yml
```diff
@@ -85,7 +88,7 @@ jobs:
           - orca-job: ISOLATED_TEST_ON_CURRENT
             php-version: "8.3"
             orca-enable-nightwatch: "TRUE"
-            orca-coverage-enable: "TRUE"
+            orca-coverage-clover-enable: "TRUE"
```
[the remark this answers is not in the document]
secretsayan: Yeah, I wanted to test if all three variables which are meant for generating coverage are actually working, so used ORCA_COVERAGE_ENABLE in php 8.1, ORCA_COVERAGE_CLOVER_ENABLE in php 8.2 and ORCA_COVERAGE_COBERTURA_ENABLE in 8.3 and verified that they are all working.

src/Domain/Tool/Phpunit/PhpUnitTask.php
secretsayan: Fixed

.github/workflows/orca.yml
```diff
@@ -147,7 +150,7 @@ jobs:
       - name: After script
         run: |
           ../orca/bin/ci/self-test/after_success.sh
-          unset ORCA_COVERAGE_ENABLE
+          unset ORCA_COVERAGE_COBERTURA_ENABLE
```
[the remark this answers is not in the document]
secretsayan: Yes, all three needs to be unset.

bin/ci/self-test/script.sh
TravisCarden: Okay, that makes sense. Well, readability is my only real concern here, so what if we just introduce a temporary variable. Something like this, maybe? (My bash is a little rusty.)

```shell
if [[ "$ORCA_COVERAGE_ENABLE" == TRUE  || "$ORCA_COVERAGE_COBERTURA_ENABLE" == TRUE  || "$ORCA_COVERAGE_CLOVER_ENABLE" == TRUE ]]; then
  ANY_COVERAGE_IS_ENABLED=TRUE
else
  ANY_COVERAGE_IS_ENABLED=TRUE
fi
```

secretsayan: I have created this` ORCA_ANY_COVERAGE_IS_ENABLED` and neither added it to any documentation and nor added it to services to keep this under the hood.

src/Enum/EnvVarEnum.php
```diff
@@ -38,6 +39,12 @@ class EnvVarEnum extends Enum {
 
   public const ORCA_COVERAGE_CLOVER = 'ORCA_COVERAGE_CLOVER';
 
+  public const ORCA_COVERAGE_COBERTURA = 'ORCA_COVERAGE_COBERTURA';
+
+  public const ORCA_COVERAGE_COBERTURA_ENABLE = 'ORCA_COVERAGE_COBERTURA_ENABLE';
```
TravisCarden, line 44: This is missing its `@method` annotation above.

TravisCarden (changes requested):
Okay, @secretsayan. Two changes, and then you may merge.

PR #503, opened 2024-02-07
Gitlab directory name mismatch fix

secretsayan:
When GitlabCI runs against an MR (created from a forked repository of the module), the CI fails. This is due to the fact that the ORCA_SUT_DIRECTORY environment (i.e $CI_PROJECT_DIR ) resolves to /builds/issue/module_name-issue-number, whereas ORCA anticipates that the last path name should be the module name, like /builds/issue/module_name. 

This PR ensures that ORCA searches for the SUT in the correct place even if its directory name does not match the actual SUT name.

### DIFF
--- a/src/Domain/Package/PackageManager.php
+++ b/src/Domain/Package/PackageManager.php
@@ -283,8 +283,7 @@ class PackageManager {
   /**
    * Checks if a package is null.
    */
-  private function containsValidVersion($data) : bool {
-
+  private function containsValidVersion($data): bool {
     if (!is_array($data)) {
       return FALSE;
     }
@@ -300,11 +299,8 @@ class PackageManager {
    * Adds a package to the list of packages.
    */
   private function addPackage(array $datum, FixturePathHandler $fixture_path_handler, string $package_name): void {
-    $orca_sut_dir = $this->env->get('ORCA_SUT_DIR');
-    if (empty($datum['url']) && !is_null($orca_sut_dir)) {
-      $orca_sut_dir = $this->env->get('ORCA_SUT_DIR');
-      $package_name_parts = explode('/', $orca_sut_dir);
-      $datum['url'] = "../" . end($package_name_parts);
+    if ($this->env->get('ORCA_SUT_NAME') === $package_name) {
+      $datum = $this->setSutUrl($datum);
     }
     $package = new Package($datum, $fixture_path_handler, $this->orca, $package_name);
     $this->packages[$package_name] = $package;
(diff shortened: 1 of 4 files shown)


secretsayan:
@TravisCarden : This is ready for review and pushed this changes to `wip` branch and confirmed with Vishal that its working fine.

wimleers:
The title says "Gitlab fix", with no description.

_What_ is this fixing?

src/Domain/Package/PackageManager.php
```diff
@@ -284,6 +300,12 @@ private function containsValidVersion($data) : bool {
    * Adds a package to the list of packages.
    */
   private function addPackage(array $datum, FixturePathHandler $fixture_path_handler, string $package_name): void {
+    $orca_sut_dir = $this->env->get('ORCA_SUT_DIR');
+    if (empty($datum['url']) && !is_null($orca_sut_dir)) {
+      $orca_sut_dir = $this->env->get('ORCA_SUT_DIR');
+      $package_name_parts = explode('/', $orca_sut_dir);
+      $datum['url'] = "../" . end($package_name_parts);
+    }
```
TravisCarden, line 308: This method is called on in `::initializePackages()` above on every package in `packages.yml`, which I believe means this would change the `url` of _all_ packages to the SUT dir. Obviously, that's not right--this needs to be done somewhere else that targets just the SUT. When you move it, why don't you add some test coverage that would've caught this.

TravisCarden (changes requested):

secretsayan:
@TravisCarden : Updated the PR and also formatted existing code.

src/Domain/Package/PackageManager.php
```diff
@@ -205,7 +216,12 @@ public function getAlterData(): array {
   private function initializePackages(FixturePathHandler $fixture_path_handler, string $packages_config, ?string $packages_config_alter): void {
     $data = $this->parseYamlFile($this->orca->getPath($packages_config));
     if ($packages_config_alter) {
-      $alter_path = $this->orca->getPath($packages_config_alter);
+      if (!$this->filesystem->isAbsolutePath($packages_config_alter)) {
+        $alter_path = $this->orca->getPath($packages_config_alter);
+      }
+      else {
+        $alter_path = $packages_config_alter;
+      }
```
TravisCarden, line 224: Just for readability, let's reverse this condition: (from "if not then don't" to "if then")

```suggestion
      if ($this->filesystem->isAbsolutePath($packages_config_alter)) {
        $alter_path = $packages_config_alter;
      }
      else {
        $alter_path = $this->orca->getPath($packages_config_alter);
      }
```

tests/Domain/Package/PackageManagerTest.php
```diff
@@ -171,7 +188,8 @@ public function testConstructionAndGetters(): void {
     self::assertEquals('drupal/module2', $package->getPackageName(), 'Got package by name.');
     self::assertEquals(self::EXPECTED_DEPENDENCY_LIST, $actual_dependency_list, 'Set/got all dependencies.');
     self::assertEquals(TRUE, $package->isCompanyPackage(), 'Got a company package.');
-    self::assertEquals(FALSE, $manager->get('drupal/dependency1')->isCompanyPackage(), 'Got a third party dependency.');
+    self::assertEquals(FALSE, $manager->get('drupal/dependency1')
+      ->isCompanyPackage(), 'Got a third party dependency.');
```
TravisCarden, line 192: Non-meaningful (whitespace only) change.

```suggestion
    self::assertEquals(FALSE, $manager->get('drupal/dependency1')->isCompanyPackage(), 'Got a third party dependency.');
```

TravisCarden (approved):
Great. One little suggestion, and then I approve.

PR #505, opened 2024-02-09
Add Acquia CMS Common module

secretsayan:
ACMS is failing the PREV_MINOR tests as ACMS Common module requires version` 3.2.x` for Drupal `10.1.x`, so adding this module in `packages.yml `with the correct version constraints.

### DIFF
--- a/src/Domain/Tool/Phpunit/PhpUnitTask.php
+++ b/src/Domain/Tool/Phpunit/PhpUnitTask.php
@@ -180,7 +180,6 @@ class PhpUnitTask extends TestFrameworkBase {
       $include = $this->doc->createElement('include');
 
       foreach ($suffixes as $suffix) {
-        echo $suffix;
         $directory = $this->doc->createElement('directory', $this->getPath());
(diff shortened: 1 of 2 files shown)


secretsayan:
@TravisCarden : Please check this out, and let me know do we need to add this module.

src/Domain/Tool/Phpunit/PhpUnitTask.php
```diff
@@ -180,6 +180,7 @@ private function setCoverageFilter(): void {
       $include = $this->doc->createElement('include');
 
       foreach ($suffixes as $suffix) {
+        echo $suffix;
```
TravisCarden, line 183: Looks like a forgotten debug statement.

TravisCarden (changes requested):
Yes, let's do this. I believe in actual use this module is placed by the ACMS starter command. But until we're testing with that, let's at least make sure the module can be installed in the codebase. I approve after the below little fix. (You can go right ahead and merge it without asking for my review again.)

PR #557, opened 2024-05-22
D11 compatibility updates 3 

secretsayan:
- Modify phpunit.xml as per D11 changes
- Update ubuntu version to support D11


TravisCarden (approved):
Looks good to me!

PR #558, opened 2024-05-23
Fix static analysis tests

secretsayan:


TravisCarden:
You definitely don't need my review on something this simple, @secretsayan. In a case like this, if the tests pass, go ahead. 🙂

secretsayan:
> You definitely don't need my review on something this simple, @secretsayan. In a case like this, if the tests pass, go ahead. 🙂

I wanted to show you, how `phpunit` 10 is breaking our simplest tests :(

secretsayan:
> Okay. Then is there something you need from me?

If you don't have any comments/suggestions, I will just merge it.

TravisCarden:
No, no concerns. In fact, it seems better this way. I assume static analysis complained because of duplicate keys, which are a genuine issue to be fixed. Also, the arrays now reflect the names of the test method arguments. So it seems better all around.

secretsayan:
> No, no concerns. In fact, it seems better this way. I assume static analysis complained because of duplicate keys, which are a genuine issue to be fixed. Also, the arrays now reflect the names of the test method arguments. So it seems better all around.

Okay, then merging

PR #599, opened 2024-07-29
Add ORCA_PHPUNIT_DEBUG_MODE_ENABLE variable

secretsayan:


TravisCarden (changes requested):
Seems good to me. I just have a couple of code-level comments. At the high level, I would point out that this is changing the default behavior, so you'll want to put out a change notice.

vishalkhode1:
@secretsayan Looks good to me. Agree with @TravisCarden here, but for D11 absolutely I'm pretty sure developers don't want to see a lot of un-related messages as it becomes quite difficult to find the source of error. :D 
> At the high level, I would point out that this is changing the default behavior, so you'll want to put out a change notice.

vishalkhode1 (approved):

PR #614, opened 2024-09-19
remove memcache-settings from ORCA as it is abandoned

Raksha-Bharuka:


Raksha-Bharuka:
This is not required anymore as memcache-settings is un-archived.

PR #663, opened 2025-04-11
Wip

mayur-sose:


Raksha-Bharuka (commented):
We should also add INTEGRATED TEST with PHP8.4

